PR #1, opened 2021-03-17
Genesis

cbolles:
# Welcome
Welcome to the initial development PR for the new EVT-Core library. This marks a new beginning for EVT firmware development. The code included in this PR will lay the foundation for the core logic used by EVT members developing for our electric vehicles. This step represents our final push to move off of Mbed to our own platform. Most of the adapters written around the HAL in the previous code base will be moved into this one.

## What is the Goal of This PR?
This PR does not represent the first full release of EVT-Core, rather it is a stepping stone towards that goal. Through the project you will find `TODO` blocks and other such comments as well as incomplete code. That is to be expected. Right now the goal is to have a working proof of concept that can compile code that is capable of running on an STM microcontroller. Below is the list of features in this PR.

* CMake based compilation process
* CMake wrapper for the STMF3xx HAL
* Basic Blink Sample
* LED device layer adapter
* GPIO IO layer adapter
* General project structure for future development

This new code base is hopefully going to be the foundation of EVT firmware development for years to come, as such it is important to raise any concerns you may have.

## What is Not Included in This PR?
As mentioned this PR is not the initial release, just a stepping stone. The code is not yet feature complete in any way to full replace the existing code base. Some of the notable absent features are as follow.

* Project wide linting
* STM32F446RE support
* Most IO adapters
* Full README
* Documentation generation for library
* CMake functions to generate bin, hex, map, and elf files

### NOTES
* When analyzing the source, all code in the HALf3 library is copy and pasted from a STM generated project and modifications will not be made without extremely good reasoning.


README.md
```diff
@@ -0,0 +1,90 @@
...
+the needed tools to generate the code.
+
+After that, the compilation process follows standard practice. Below is
+an example compilation process.
+
+```bash
+mkdir build/
+cd build/
+cmake ../
+make
```
jjenco97, line 46: For a parallel build:
```suggestion
make -j $(nproc)
```

jjenco97:
General comment on namespaces: I only see `EVT::core` as a means to access lower-level namespaces. Is there anything you plan on using those two namespaces for? I do like how it mostly matches the directory structure, but typing out 5 levels of scope to get at an enum type is a little much with a project of this stature. You seem to make this point for yourself in lines 14-16 of the blink sample code. (`namespace IO = ...`)

My suggestion: The directory structure `EVT-Core/include/EVT/**/` is a little redundant, perhaps remove `EVT` and prune the namespaces to match. This is already the case with the `src/` directory. We already know this is an EVT project, if you want to make a distinction with 3rd party libs, you could devote a subdirectory (and namespace) for that later on.

jjenco97:
Now is probably the best time to have a discussion about the current coding standards, since the last time that happened was when we did an overhaul of this magnitude. (For example, file names like `GPIOf3.cpp` should be more like `gpio_f3.cpp` as the spec is currently written)

README.md
ssnover: For a build machine it's probably fine to just use -j to allow make to decide.

cbolles:
> Now is probably the best time to have a discussion about the current coding standards, since the last time that happened was when we did an overhaul of this magnitude. (For example, file names like `GPIOf3.cpp` should be more like `gpio_f3.cpp` as the spec is currently written)

Yes I plan on the first release to have an enforced style guide. I think the convention of matching the name of the file to the contained class is advantageous since it makes searching for code in a repo much easier.

cbolles:
> General comment on namespaces: I only see `EVT::core` as a means to access lower-level namespaces. Is there anything you plan on using those two namespaces for? I do like how it mostly matches the directory structure, but typing out 5 levels of scope to get at an enum type is a little much with a project of this stature. You seem to make this point for yourself in lines 14-16 of the blink sample code. (`namespace IO = ...`)
> 
> My suggestion: The directory structure `EVT-Core/include/EVT/**/` is a little redundant, perhaps remove `EVT` and prune the namespaces to match. This is already the case with the `src/` directory. We already know this is an EVT project, if you want to make a distinction with 3rd party libs, you could devote a subdirectory (and namespace) for that later on.

The main goal of having the distinction of EVT::core is to separate the core library functionality from the future board code. So for example, the will be EVT::TMS, EVT::BMS, etc for the board specific code. By the way with the current architecture those will be in different repositories.

The folder convention of `/include/<project_name>` is actual a common convention for library code which other projects use.
ex) [Opencv](https://github.com/opencv/opencv), each sub project in Mbed follows that convention.
In the future when we have the board repositories, having the extra `EVT` include will make determining where source code is coming from easier.

jjenco97:
> > General comment on namespaces: I only see `EVT::core` as a means to access lower-level namespaces. Is there anything you plan on using those two namespaces for? I do like how it mostly matches the directory structure, but typing out 5 levels of scope to get at an enum type is a little much with a project of this stature. You seem to make this point for yourself in lines 14-16 of the blink sample code. (`namespace IO = ...`)
> > My suggestion: The directory structure `EVT-Core/include/EVT/**/` is a little redundant, perhaps remove `EVT` and prune the namespaces to match. This is already the case with the `src/` directory. We already know this is an EVT project, if you want to make a distinction with 3rd party libs, you could devote a subdirectory (and namespace) for that later on.
> 
> The main goal of having the distinction of EVT::core is to separate the core library functionality from the future board code. So for example, the will be EVT::TMS, EVT::BMS, etc for the board specific code. By the way with the current architecture those will be in different repositories.
> 
> The folder convention of `/include/<project_name>` is actual a common convention for library code which other projects use.
> ex) [Opencv](https://github.com/opencv/opencv), each sub project in Mbed follows that convention.
> In the future when we have the board repositories, having the extra `EVT` include will make determining where source code is coming from easier.

Ahh okay, that makes much more sense when all the pieces of the puzzle are there. I can get behind it. Well done.

README.md
```diff
@@ -0,0 +1,90 @@
+# EVT-core
+
+## Introduction
+
+### What is EVT?
+EVT stands for the Electric Vehicle Team at RIT. We build and race electric
+vehicles. Currently we have two track motorcycles we develop for and are
+desiging an electric dirt bike.
```
ddp5730, line 8: spelling

README.md
```diff
@@ -0,0 +1,90 @@
...
+
+```bash
+mkdir build/
+cd build/
+cmake ../
+make -j
+```
+
+After that, in the `build` folder will be the compiled library code as well
+as generated samples.
```
ddp5730, line 50: Should we link to the Wiki page with board programming instructions?
Also should we create a new page with instructions of how to set up build/programming configurations within Clion?

README.md
cbolles: Nice catch

README.md
cbolles: I think in the future we may want to link to additional instructions. I was planning on finishing the README once the rest of the system is set so the information is not outdated.

ddp5730: Sounds good.

include/EVT/io/pin.hpp
```diff
@@ -0,0 +1,109 @@
...
+#define _EVT_PIN_
+
+namespace EVT::core::IO
+{
+// STM32F302R8 pin names
+#ifdef STM32F302x8
+
+    enum class Pin
+    {
+        PA_0  = 0x00,
```
ddp5730, line 16: Could we add some comment or documentation pointing to the definition for the mappings in this file?

include/EVT/utils/bits.hpp
```diff
@@ -0,0 +1,79 @@
...
+// 16 bit operations
+/**
+ * Get the high byte from the number.
+ *
+ * @param data The number to get the high byte from
+ * @return The high order byte
+ */
+constexpr static inline uint8_t getHighByte(uint16_t data)
+{
+    return static_cast<uint8_t>((data & 0xFF00) >> 8);
```
ddp5730, line 20: Shouldn't function definitions be in the source file instead of header file?  Or is this normal practice for constexpr?

ddp5730 (commented):
Looks really good.  A few small comments to address before approval.

include/EVT/io/pin.hpp
cbolles: Good point, just updated.

include/EVT/utils/bits.hpp
cbolles: Since they are inline, they don't operate like normal functions. Instead of the call generating a jump statement to the function. The code is expanded inline where there is the call. That is why the code is in the header.

ddp5730 (approved):

include/EVT/io/GPIO.hpp
```diff
@@ -0,0 +1,109 @@
...
+namespace EVT::core::IO
+{
+    // The different pins are hardware specific. Forward declaration to allow
+    // at compilation time the decision of which pins should be used.
+    enum class Pin;
+}
+
+
+namespace EVT::core::IO
+{
```
jjenco97, line 16: Redundancy?
```suggestion
namespace EVT::core::IO
{
// Forward declarations:
// The different pins are hardware specific. Forward declaration to allow
// at compilation time the decision of which pins should be used.
enum class Pin;

```

src/io/GPIO.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    this->pin = pin;
+    this->direction = direction;
+}
+
+GPIO& GPIO::getInstance(Pin pin, Direction direction)
+{
+    // Determine which GPIO interface to use
+    #ifdef STM32F302x8
+        static GPIOf3 gpioPin(pin, direction);
```
jjenco97, line 29: I'm not totally a fan of labelling the GPIO adapter as just "F3", you're moving from less to more abstract here. The handlers written for it are specific to "F302x8", and may not necessarily apply to all F3 models. Same applies for the other future IO modules.

samples/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+# Add all samples
+add_subdirectory(blink)
```
jjenco97, line 2: Not an action item for this PR, but something convenient for the future - create a simple shell script that auto-generates a new CMakeLists.txt based on a template and appends the `add_subdirectory` in the appropriate location(s) every time you want to make a new executable

src/io/GPIO.cpp
cbolles: I think for right now we should be ok, the code in the GPIOf3 adapter does not necessarily make use of F302x8 specific code. In the future if we adopt more F3 models, then I think we would extend GPIOf3 to have GPIOF302x8 etc.

samples/CMakeLists.txt
cbolles: I'll look into how to do that in CMake, I shy away from making a shell script since that takes away the Cross Platformness of CMake.

src/io/GPIO.cpp
jjenco97: You may be okay with GPIO because it is so trivial, but I can see other peripherals on other F3 models where you could run into trouble here. The IRQ handlers are defined initially in `startup_stm32f302x8.s` and their layout is most definitely model-specific. And while the handler names may be the same from model to model, it's not guaranteed that their implementation will be that way. I hate to hold up a PR over a semantics argument but if we're going through all the trouble to support multiple controller families we might as well be this explicit from the get go.

cbolles: Yeah that's a good point, I added a sub-folder to f3xx that has the f302x8 code renamed within. In the future if we expand to include more f3 models they can go into the f3xx folder.

include/EVT/io/GPIO.hpp
cbolles: Yeah that is redundant. Originally I had that to separate the code from the forward declarations, but the command should be plenty.

jjenco97 (approved):

PR #4, opened 2021-06-13
CAN can now CAN

cbolles:
Adds basic CAN transmit and receive functionality. Users can transmit and receive over a CAN bus.

As opposed to the implementation in mbed-applications this implementation does not support receiving based on a specific CAN ID. This was to avoid the need to dynamic memory allocations and maps for CAN message buffers. Instead receiving CAN messages will give you the most recently received message.


include/EVT/io/CAN.hpp
```diff
@@ -0,0 +1,74 @@
...
+     * In non-blocking, a nullptr will be returned if no message is currently
+     * in the mailbox.
+     *
+     * @param message {out} The message to populate with data
+     * @param blocking Used to determine if received should block or not, by
+     *      default receive is blocking
+     * @return A pointer to the passed in message, nullptr if message not
+     *      received.
+     */
+    virtual CANMessage* receive(CANMessage* message, bool blocking=true) = 0;
```
jjenco97, line 60: Not sure I like having blocking behavior as the default...99% of the time you'll just check if something is present and then act on it, and if it's not there we move on.

include/EVT/io/CAN.hpp
```diff
@@ -0,0 +1,74 @@
...
+public:
+    /**
+     * Creates a new instance of the CAN interface which will use the given
+     * transmit and receive pins.
+     *
+     * @param txPin The pin to use for transmitting data
+     * @param rxPin The pin to use for receiving data
+     * @param loopbackEnabled Flag to enable CAN loop back functionality
+     */
+    CAN(Pin txPin, Pin rxPin, bool loopbackEnabled);
```
jjenco97, line 38: Question, since I'm not 100% on your new architecture layout - is the only way to instantiate the CAN class by means of the manager? If so this is fine, otherwise I'd say put the default option for loopbackEnabled here, and have it false.

include/EVT/utils/types/FixedQueue.hpp
```diff
@@ -0,0 +1,167 @@
...
+// The most number of elements in the queue, the user can provide a
+// value that is less then this number to reduce the apparent capacity.
+// Note that doing so will not reduce the size of the instance.
+#define QUEUE_MAX_CAPACITY 512
+
+namespace EVT::core::types
+{
+
+template<class Element>
+class FixedQueue {
```
jjenco97, line 30: Big fan of this. Reminds me of something I did back at Magna (non-CAN related)! Definitely a better way of handling messages.

src/io/types/CANMessage.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <EVT/io/types/CANMessage.hpp>
+
+namespace EVT::core::IO {
+
+CANMessage::CANMessage(uint32_t id, uint8_t dataLength, uint8_t* payload) {
+    this->id = id;
+    this->dataLength = dataLength;
+
+    // Copy contents of provided payload into message's payload
+    for (int i = 0; i < dataLength; i++)
```
jjenco97, line 13: Upper bounds check on dataLength being larger than 8 bytes?

src/io/CAN.cpp
```diff
@@ -0,0 +1,13 @@
+#include <stdint.h>
+
+#include <EVT/io/CAN.hpp>
+
+namespace EVT::core::IO {
+
+CAN::CAN(Pin txPin, Pin rxPin, bool loopbackEnabled) {
+    this->txPin = txPin;
+    this->rxPin = rxPin;
+    this->loopbackEnabled = loopbackEnabled;
```
jjenco97, line 10: Depending on your stylistic preferences, I suppose.
```suggestion
CAN::CAN(Pin txPin, Pin rxPin, bool loopbackEnabled) :
    txPin(txPin),
    rxPin(rxPin),
    loopbackEnabled(loopbackEnabled) {
```

samples/queue/main.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // Test a queue of numbers, no overwritting
+    ///////////////////////////////////////////////////////////////////////////
+    uart.printf("Queue of Numbers, no overwritting\n\r");
+    // First a queue of numbers, max capacity of 10 and no over writting
+    types::FixedQueue<int> numberQueue(10);
+    // Check if it is empty, should be empty
+    uart.printf("numberQueue.isEmpty() -> %d\r\n", numberQueue.isEmpty());
+
+    // Add ten numbers, the max allowed
+    for (int i = 0; i < 10; i++) {
```
jjenco97, line 59: A little pedantic, but you don't try to over-fill the queue in the non-overwrite case, or pop from an empty queue. Just to be complete.

jjenco97 (commented):
Overall, pretty well done, anything I've found on first glance is pretty minor.

Do you guys have the USB to isoCAN boards working? Or some other means to send messages between two devices? You could write a script to stream CAN messages from one device to another and find the point at which the message queue starts to break down. The Linux can-utils can give bus utilization characteristics as well. It would be a good stress test.

include/EVT/io/CAN.hpp
cbolles: Yeah I was on the edge with this one, but I can see after digging around that non-blocking is the usual default. I think you're right, blocking should default to false.

include/EVT/io/CAN.hpp
cbolles: No the user should be able to use the manager or construct a CAN instance directly, so I agree this should also have a default option.

include/EVT/utils/types/FixedQueue.hpp
cbolles: Thanks! I was happy with how the queue based version worked out.

src/io/types/CANMessage.cpp
cbolles: Yeah that should have bound checking. In the future we should have either an assertion or other way to notify the user of an issue, but for now I'll just bound the dataLength to the max frame size.

src/io/CAN.cpp
cbolles: I think your method is generally the preferred one, I'll switch to that

samples/queue/main.cpp
cbolles: I don't think that's too pedantic, the queue should be a known working thing. I'll add test cases for those

cbolles:
> Overall, pretty well done, anything I've found on first glance is pretty minor.
> 
> Do you guys have the USB to isoCAN boards working? Or some other means to send messages between two devices? You could write a script to stream CAN messages from one device to another and find the point at which the message queue starts to break down. The Linux can-utils can give bus utilization characteristics as well. It would be a good stress test.

Haven't had much luck with the isoCAN board, but I just got an arduino with CAN transceiver that I plan on using for testing. I'm thinking of using one of our CAN log files pulled from the GUB as an input that I give to the arduino code and the arduino will spit out the logged messages. Gives it a more "realness" feel.

jjenco97:
> > Overall, pretty well done, anything I've found on first glance is pretty minor.
> > Do you guys have the USB to isoCAN boards working? Or some other means to send messages between two devices? You could write a script to stream CAN messages from one device to another and find the point at which the message queue starts to break down. The Linux can-utils can give bus utilization characteristics as well. It would be a good stress test.
> 
> Haven't had much luck with the isoCAN board, but I just got an arduino with CAN transceiver that I plan on using for testing. I'm thinking of using one of our CAN log files pulled from the GUB as an input that I give to the arduino code and the arduino will spit out the logged messages. Gives it a more "realness" feel.

That's exactly what they do in industry for trying to replicate vehicle issues that have been isolated to one particular location. They capture every message on CAN network and replay the time sequence around the failure to try to reproduce it in the lab. (Although their setups are a bit more elaborate than an Arduino). To have something like that would be sweet.

jjenco97 (approved):

PR #5, opened 2021-08-07
Feature/cbolles/adc

cbolles:
Brings in ADC support for the F302x8. ADC  works for a single pin so far.

Note worth TODOs.

1) Only single pin ADC is supported. The second pin does not seem to give back accurate values. This is actually identical behavior to the mbed_applications implementation as well. I believe it is how we keep the DMA buffer as 16 bit numbers, but the ADC works with 32 bit numbers.

2) There is system level initialization that takes place in the ADC. This is just so the basic ADC function can be proved. It will be resolved before 0.0.2 is merged into main.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,7 @@ set(CMAKE_CXX_FLAGS         "${EVT_COMMON_FLAGS} \
 # TODO: Check for platform to decide which linker script to use
 set(CMAKE_EXE_LINKER_FLAGS  "-mfloat-abi=hard -specs=nano.specs -specs=nosys.specs \
                             -T ${PROJECT_SOURCE_DIR}/libs/HALf3/STM32F302C8Tx_FLASH.ld \
-                            -lc -lm -lnosys")
+                            -lc -lm -lnosys -u _printf_float")
 
 ###############################################################################
 # Handle dependencies
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,10 +97,12 @@ add_library(${PROJECT_NAME} STATIC)
 
 # Add sources
 target_sources(${PROJECT_NAME} PRIVATE
+    src/io/ADC.cpp
     src/io/CAN.cpp
     src/io/GPIO.cpp
     src/io/PWM.cpp
     src/io/UART.cpp
+    src/io/platform/f3xx/f302x8/ADCf302x8.cpp
     src/io/platform/f3xx/f302x8/CANf302x8.cpp
     src/io/platform/f3xx/f302x8/GPIOf302x8.cpp
     src/io/platform/f3xx/f302x8/PWMf302x8.cpp
--- a/include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
+++ b/include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
@@ -56,7 +56,7 @@ private:
     static Pin channels[MAX_CHANNELS];
     /// Buffer for DMA where each spot represents the value read in from a
     /// channel
-    static uint32_t buffer[MAX_CHANNELS];
+    static uint16_t buffer[MAX_CHANNELS];
     static DMA_HandleTypeDef halDMA;
 
     /**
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -133,7 +135,7 @@ set(CMAKE_CXX_FLAGS         "${EVT_COMMON_FLAGS} \
 # TODO: Check for platform to decide which linker script to use
 set(CMAKE_EXE_LINKER_FLAGS  "-mfloat-abi=hard -specs=nano.specs -specs=nosys.specs \
                             -T ${PROJECT_SOURCE_DIR}/libs/HALf3/STM32F302C8Tx_FLASH.ld \
-                            -lc -lm -lnosys")
+                            -lc -lm -lnosys -u _printf_float")
```
jjenco97, line 138: Be careful that we don't abuse this, it's disabled by default for a reason. For ADC readings, multiplying by 1000 and static casting to an integer keeps your precision without the overhead for pretty printouts, both from a code size and execution time perspective.

include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * @return The ADC value as a percentage
+     */
+    float readPercentage();
+
+
+private:
+    // Max number of channels supported by the ADC
+    static constexpr uint8_t MAX_CHANNELS = 15;
+    // Max voltage of the ADC
+    static constexpr float MAX_VOLTAGE = 3.6;
```
jjenco97, line 48: Curious this is 3.6? I would think 3.3, but if the datasheet says otherwise, obviously keep it as is.

CMakeLists.txt
cbolles: Oh great point. In the future u can move this to the compile flags of the specific sample

include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
cbolles: Yes I found this somewhere as the max, i can find the source if you would like. I think its based on the max logic high level the system has in spec

src/io/platform/f3xx/f302x8/ADCf302x8.cpp
```diff
@@ -0,0 +1,252 @@
...
+namespace EVT::core::IO {
+
+// Init static member variables
+ADC_HandleTypeDef ADCf302x8::halADC = {0};
+Pin ADCf302x8::channels[MAX_CHANNELS];
+uint16_t ADCf302x8::buffer[MAX_CHANNELS];
+DMA_HandleTypeDef ADCf302x8::halDMA = {0};
+
+
+ADCf302x8::ADCf302x8(Pin pin) : ADC(pin) {
```
jjenco97, line 70: Constructor delegation? I don't think I've ever seen this in practice before.

cbolles: Yep!

include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
jjenco97: Page 293 of the reference manual seems to have it.

src/io/platform/f3xx/f302x8/ADCf302x8.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+extern "C" void DMA1_Channel1_IRQHandler(void) {
+    HAL_DMA_IRQHandler(dmaHandle);
+}
+
+/**
+ * Current "work around" solution that ensurs the clock settings are applied
+ * and the HAL is initialized. This will later be moved into a common place
+ * that will happen once at system start up. For the time being this can be
+ * used for basic setup. This is not a long term solution.
```
jjenco97, line 26: This better not be a long term solution :joy:

cbolles: Oh god no. After I2C im going to determine the best way for a system init. This is trash right now

jjenco97: If the ```feature/system_clock_init``` branch still exists in the old repo, feel free to copy-paste those clock settings for each target. If the designs of the boards haven't changed dramatically (don't see why they'd need to) then those values should still be valid. Otherwise you can generate them yourself with STM32CubeMX.

jjenco97 (approved):

zsteig (approved):

PR #6, opened 2021-08-12
I too see a new feature

cbolles:
I2C brought in from mbed-applications. Adds additional wrapper methods for common I2C interactions (reading from registers, reading from registers with addresses larger then 8 bits, etc). Still a little rough around the edges, but good starting point.


src/io/I2C.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * Contains generic implmentations for some of the I2C funcitonality.
+ * Some hardware has the ability to implement the functionality in this
+ * file so specific implementations of the I2C driver may override these
+ * methods for a hardware speedup.
+ */
+#include <EVT/io/I2C.hpp>
+
+namespace EVT::core::IO {
+
+    I2C::I2C(Pin sclPin, Pin sdaPin): sclPin(sclPin), sdaPin(sdaPin) {}
```
jjenco97, line 11: Not sure if this was present in previous PRs but function definitions inside a namespace shouldn't be indented.

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+                gpioInit.Alternate = GPIO_AF2_I2C3;
+                break;
+            default:
+                break;
+        }
+
+        Pin i2cPins[2]   = {sclPin, sdaPin};
+
+        gpioInit.Pin = static_cast<uint32_t>(1
```
jjenco97, line 72: Cases like this are where the indentation comment I made above will save you a bit. Also, are we enforcing an 80-char line width? I feel like in 2021 that's really something we don't have to worry about. Lines 72, 121, 127, etc. look kinda ugly to read.

include/EVT/io/I2C.hpp
```diff
@@ -0,0 +1,88 @@
...
+    void read(uint8_t addr, uint8_t* bytes, uint8_t length);
+
+    /**
+     * Read a value from a register that has an 8 bit address and 8 bit value
+     *
+     * @param addr The 7 bit unshifted I2C address to read from.
+     * @param reg The 8 bit register
+     * @return The 8 bit value of the register.
+     */
+    uint8_t readReg(uint8_t addr, uint8_t reg);
```
jjenco97, line 65: readReg but no writeReg?

src/io/I2C.cpp
cbolles: Woops, yeah that should not be indented. I'll change that

include/EVT/io/I2C.hpp
cbolles: Yeah there should be a write reg, I'll add that now

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
jjenco97: Not anymore, technically https://www.phoronix.com/scan.php?page=news_item&px=Linux-Kernel-Deprecates-80-Col

cbolles: Ah very cool, didn't realize they had lengthened the limit. Then I think it could be worth changing it in the future.

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
```diff
@@ -0,0 +1,150 @@
+#include <EVT/io/pin.hpp>
+#include <EVT/io/platform/f3xx/f302x8/I2Cf302x8.hpp>
+
+namespace EVT::core::IO {
+
+    /**
```
jjenco97, line 6: Same issue here with premature indents

cbolles: I've become too reliant on my out-indent it would seem

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
```diff
@@ -0,0 +1,150 @@
...
+            return 1;
+        case Pin::PA_9:
+        case Pin::PF_1:
+            return 2;
+        case Pin::PA_8:
+            return 3;
+        default:
+            return 0;
+    }
+    return 0;  // Should not get here, unless bad pin given
```
jjenco97, line 32: What actually happens when you give a bad pin? In the best case, nothing, but in the worst case, maybe a hard fault? You still reach ```HAL_I2C_Init()``` regardless of whether or not you enable the clocks and do everything properly.

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+namespace EVT::core::IO {
+
+/**
+ * Get the I2C ID associated that is associated with the given pin
+ * combination. This information is pulled from the STM32F302x8
+ * documentation with easier documentation on MBed's STM32F302r8
+ * documentation.
+ *
+ * TODO: Make modifications for this to be runnable at compile time
```
jjenco97, line 12: The old repo has this done for SPI if I remember correctly. Perhaps you could use it as reference.

include/EVT/io/platform/f3xx/f302x8/I2Cf302x8.hpp
```diff
@@ -0,0 +1,108 @@
...
+     * @param regLength The size in bytes of the register
+     * @param bytes The bytes read from the register
+     * @param length The size of the data returned by the register in bytes
+     */
+    void readReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
+            uint8_t* bytes, uint8_t length);
+
+private:
+    constexpr static uint32_t DEFAULT_I2C_FREQ = 100000;
+    constexpr static uint32_t DEFAULT_I2C_TIMEOUT = 100;
```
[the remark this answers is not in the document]
cbolles: Honestly not sure, this was grabbed from mbed-applications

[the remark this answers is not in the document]
cbolles: Honestly not sure, this was grabbed from mbed-applications

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
cbolles: This will hopefully be part of a compile time check. I also plan on implementing actual error handling in the future. We should be checking the return from every call, including HAL calls for errors

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
cbolles: Cool, I'll look to pull that into the future. Right now im focusing on bare minimum functionality, and over the semester working on improvements

include/EVT/io/platform/f3xx/f302x8/I2Cf302x8.hpp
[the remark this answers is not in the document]
jjenco97: On a deeper dive it looks like it's based on the number of system ticks, which is a global variable deep in the HAL used to maintain system time. Default is 1ms, unless we change it at all.

```HAL_I2C_Master_Transmit() --> I2C_WaitOnTXISFlagUntilTimeout() --> HAL_GetTick()```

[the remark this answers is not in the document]
cbolles: Should still me 1ms, its what we use for "waiting". I can add a comment later tonight

jjenco97 (approved):

PR #7, opened 2021-08-23
Release the code

cbolles:
We are ready for our first "official release" 0.1.0 which is currently based off of `dev/0.0.2` will include everything from `dev/0.0.2` as well as the clean up commits I have made. The idea is that EVT-core is now feature complete enough to start writing various board in EVT-core. v0.1.0 will be the starting point from which to start development.

We will continue to make tweaks until EVT-core is considered "stable" (no longer having large scale changes), At which point we will have our 1.0.0 release.

The main things added in the cleanup are

1. A macro for creating an executable that  can be used to flash a board
2. Creation of a system level init function which handles initializing the board including clock settings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,8 @@ add_library(${PROJECT_NAME} STATIC)
 
 # Add sources
 target_sources(${PROJECT_NAME} PRIVATE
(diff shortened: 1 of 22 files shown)


ddp5730 (approved):
Looks good to me.

PR #9, opened 2021-08-29
Add ability to test back and forth CAN communication

cbolles:
Added sample of CAN message sending back and forth between two devices. An Arduino example was provided for testing with an Arduino with a CAN shield.


samples/can/back_and_forth/arduino_can/arduino_can.ino
```diff
@@ -0,0 +1,78 @@
...
+    Serial.println();
+  }
+
+  Serial.println();
+
+  readyToSend = true;
+}
+
+void setup() {
+  Serial.begin(9600);
```
ddp5730, line 45: In general, I like to see important hardcoded numbers at the top of a file in a constant for easy reference.

ddp5730 (approved):
One small nitpick, but otherwise looks great!

PR #10, opened 2021-08-29
Bug/cbolles/uart receive

cbolles:
Fix issue with UART TX + RX. The issue turned out to be a couple of things.

1. The UART needs the `SysTick_Handler` function to exist, but it is optimized out unless `EVT::core::time::wait` is called (this should be resolved)
2. The `getc` method was not actually blocking so was timing out and returning garbage data. The method is now blocking, but in the future a non-blocking variant should be added.
3. With the addition of the ADC the clock settings needed to be tweaked slightly

### DIFF
--- a/src/io/platform/f3xx/f302x8/UARTf302x8.cpp
+++ b/src/io/platform/f3xx/f302x8/UARTf302x8.cpp
@@ -152,7 +152,7 @@ void UARTf302x8::puts(const char* s) {
 
 char UARTf302x8::getc() {
     uint8_t c;
-    HAL_UART_Receive(&halUART, &c, 1, DEFAULT_TIMEOUT);
+    while(HAL_UART_Receive(&halUART, &c, 1, DEFAULT_TIMEOUT) == HAL_TIMEOUT);
     return static_cast<char>(c);
 }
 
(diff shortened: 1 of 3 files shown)


ddp5730:
cpplint returned some formatting suggestions on this branch.

src/io/platform/f3xx/f302x8/UARTf302x8.cpp
```diff
@@ -152,7 +152,7 @@ void UARTf302x8::puts(const char* s) {
 
 char UARTf302x8::getc() {
     uint8_t c;
-    HAL_UART_Receive(&halUART, &c, 1, DEFAULT_TIMEOUT);
+    while(HAL_UART_Receive(&halUART, &c, 1, DEFAULT_TIMEOUT) == HAL_TIMEOUT);
```
ddp5730, line 155: The DEFAULT_TIMEOUT is pretty low.  It doesn't make much difference, but do we want to increase the timeout to this getc call as well?

ddp5730 (approved):
Looks good, I'd just like to see the cpplint warnings addressed prior to merge.

src/io/platform/f3xx/f302x8/UARTf302x8.cpp
cbolles: I think the best approach will be to later make this a parameter with a default option so that users can specify what it is. Since right now this is blocking anyway I think we are fine having a smaller timeout.

cbolles:
> Looks good, I'd just like to see the cpplint warnings addressed prior to merge.

Good catch, just fixed those issues

PR #14, opened 2021-09-02
Improve Windows build support

cbolles:
Improve windows support. Add ".exe" suffix to compilation tools on windows systems. Improve Windows build documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,30 @@ to learn about WSL (Windows Subsystem for Linux). Linux is the OS of choice
 for software developers and includes a lot of great tools for development. WSL
 is a happy medium between running Windows and installing a virtual machine.
 
+Below are incomplete and experimental instructions for building on Windows. Using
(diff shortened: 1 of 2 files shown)


jjenco97 (approved):
If your definition of improvement is to provide "something" over "nothing" then yes, I'm on board with this :joy:. At least you're honest in your documentation saying "incomplete and experimental". Hopefully this will coax enough folks into using Linux; a life skill that every software developer should have IMO.

cbolles:
> If your definition of improvement is to provide "something" over "nothing" then yes, I'm on board with this joy. At least you're honest in your documentation saying "incomplete and experimental". Hopefully this will coax enough folks into using Linux; a life skill that every software developer should have IMO.

Haha yeah, it is very much "incomplete at best". We have much better CLion support documentation (thanks Dan) and CLion will handle battling the windows build process for us. So if users really want to struggle with native CMake on windows they can

PR #15, opened 2021-09-10
RTC

cbolles:
Bring in RTC support as a device of the STM32f302r8. Includes getting a struct back as a representation of time as well as the seconds since epoch.


include/EVT/dev/platform/f3xx/f302x8/RTC302x8.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Empty constructor. Will startup the RTC.
+     */
+    RTCf302x8();
+
+    /**
+     * Get the current time as determined by the real time clock.
+     *
+     * @param[out] time The time struct to populate
```
ddp5730, line 28: I haven't seen this `[out]` or `[in]` syntax before.  Is this used by sphinx?

src/dev/platform/f3xx/f302x8/RTCf302x8.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    uint32_t y = ts.year;
+    uint32_t m = ts.month;
+    uint32_t d = ts.day;
+    uint32_t h = ts.hour;
+    uint32_t min = ts.minute;
+    uint32_t s = ts.second;
+
+
+    // Get number of days sinc epoch
```
ddp5730, line 70: Typo

src/dev/platform/f3xx/f302x8/RTCf302x8.cpp
```diff
@@ -0,0 +1,108 @@
...
+    y -= m <= 2;
+    const uint32_t era = (y >= 0 ? y : y-399) / 400;
+    const uint32_t yoe = static_cast<uint32_t>(y - era * 400);
+    const uint32_t doy = (153*(m + (m > 2 ? -3 : 9)) + 2)/5 + d-1;
+    const uint32_t doe = yoe * 365 + yoe/4 - yoe/100 + doy;
+    uint32_t time = era * 146097 + static_cast<int32_t>(doe) - 719468;
```
ddp5730, line 76: Ngl, I have no clue at all what is happening here.

ddp5730 (changes requested):
It looks great.  I'd just like to clear up why the implementation of RTCf302x8::getTime is so convoluted.  The implementation just isn't very readable or intuitive.

include/EVT/dev/platform/f3xx/f302x8/RTC302x8.hpp
cbolles: Its actually a C/C++ doc convention. In means the parameter is used as output. Out means the parameter is being used as an output. So in this example the time struct is actually storing the output of this function.

src/dev/platform/f3xx/f302x8/RTCf302x8.cpp
cbolles: Good catch, should be fixed now

src/dev/platform/f3xx/f302x8/RTCf302x8.cpp
cbolles: That's fair. Its a strange algorithm that I don't fully understand either. I've linked to the original source. I did verify that the algorithm does indeed work by comparing the output to a known good converter.

cbolles: That's fair. Its a strange algorithm that I don't fully understand either. I've linked to the original source. I did verify that the algorithm does indeed work by comparing the output to a known good converter.

include/EVT/dev/platform/f3xx/f302x8/RTC302x8.hpp
zsteig: > Its actually a C/C++ doc convention.

Will we be adopting this convention to the rest of the codebase?

include/EVT/utils/time.hpp
```diff
@@ -20,5 +20,35 @@ namespace EVT::core::time
...
+    /**
+     * Struct representing a timestamp.
+     */
+    struct TimeStamp {
+        /**
+         * The current year
+         */
+        uint16_t year;
+        /**
+         * The current month (0-12)
```
zsteig, line 33: Shouldn't this be (1-12)? Or (0-11)?

include/EVT/dev/platform/f3xx/f302x8/RTC302x8.hpp
cbolles: Ideally yes, just to clear up if an argument is used as an input, or if the caller should expect the parameter is store the output of the call.

include/EVT/utils/time.hpp
cbolles: Lol yes, good catch, 1-12. I'll fix that now

src/dev/platform/f3xx/f302x8/RTCf302x8.cpp
ddp5730: Fair enough.  I just wanted to call attention to it.

ddp5730 (approved):

zsteig (approved):

PR #16, opened 2021-09-18
READ THE DOCS

cbolles:
Bring in Sphinx documentation support into EVT-core. This adds the ability to generate documentation from the code as well as RST files. For an example hosting, [check this out](https://evt-core.readthedocs.io/en/latest/index.html). Once this is merged into main, the branch to store the latest changes will update to main.

The documentation included is far from exhaustive, this PR is intended more for bringing in the functionality of the documentation.


.readthedocs.yaml
```diff
@@ -0,0 +1,16 @@
...
+    configuration: docs/source/conf.py
+
+build:
+    image: latest
+    apt_packages:
+        - doxygen
+
+# Explicitly set the version of Python and its requirements
+python:
+    version: 3.7
```
ddp5730, line 14: Do we want to explicitly set python version?  On my machine I'm running 3.8 and I'm not sure if that would cause issues.

docs/requirements.txt
```diff
@@ -0,0 +1,30 @@
...
+snowballstemmer==2.1.0
+Sphinx==4.1.2
+sphinx-rtd-theme==0.5.2
+sphinxcontrib-applehelp==1.0.2
+sphinxcontrib-devhelp==1.0.2
+sphinxcontrib-htmlhelp==2.0.0
+sphinxcontrib-jsmath==1.0.1
+sphinxcontrib-qthelp==1.0.3
+sphinxcontrib-serializinghtml==1.1.5
+urllib3==1.26.6
```
ddp5730, line 30: For these requirements do we really want to dictate the specific versions?  In my experience we should at least make all of these `>=`, but we may want to omit the version number unless we have a specific requirement.

docs/source/conf.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'EVT-core'
+copyright = '2021, EVT'
+author = 'EVT'
+
+# The full version, including alpha/beta/rc tags
+release = '0.1.2'
```
ddp5730, line 25: Could we set the version in CmakeLists.txt as an environment variable and use that here to prevent two updates being necessary?

ddp5730 (commented):
Why is there an empty file `docs/.nojekyll` contained in this PR?

In general, looks good but I left some comments.  Looking at the webpage, I think we'll want to add some pages for general documentation (flowcharts, communication interfaces, etc) when we expand this to specific projects.  But for just EVT-core this looks good.

docs/source/conf.py
cbolles: Ahh good catch, meant to address this. A common way I've seen is to have a file with the version so that different tools know where to look for the version. I've just updated the system (both docs and CMakeLists.txt) to read in version from a file.

docs/requirements.txt
cbolles: This is the standard output from `pip freeze > requirements.txt` it's mostly a stability thing as all the build tools (including the service that hosts the documentation) will manually install these requirements before building, so this makes sure they always have a working setup.

.readthedocs.yaml
cbolles: This is purely for read the docs which is the service that hosts the website, so no one will be impacted by the python version.

cbolles:
> Why is there an empty file `docs/.nojekyll` contained in this PR?
> 
> In general, looks good but I left some comments. Looking at the webpage, I think we'll want to add some pages for general documentation (flowcharts, communication interfaces, etc) when we expand this to specific projects. But for just EVT-core this looks good.

Nice, yeah the .nojekyll was left over from attempting to host with GitHub pages, an empty file named that is needed so that GitHub doesn't apply it's own themes. Now that we are hosted with RTD, this isn't needed so I will remove it

docs/requirements.txt
ddp5730: I understand why you'd want all versions defined to have a stable build version.  But in a year from now it's entirely possible that we may be requiring all users to have very outdated packages.  Furthermore, I'm not convinced we would have any dependency issues,

ddp5730 (approved):
Personally I still might implement a less strict requirements.txt.  However, I'm fine with whatever decision you make.   Everything else looks great!

PR #18, opened 2021-09-25
Feature/dpopp/timer driver

ddp5730:
- Implemented the Timer Driver
- Timer triggers interrupt with given period
- Can start, stop, and reload the timer.


src/dev/platform/f3xx/f302x8/Timerf302x8.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+void Timerf302x8::startTimer(void (*irqHandler)(TIM_HandleTypeDef *htim)) {
+    TIM_TypeDef *timerPeripheral = this->halTimer->Instance;
+    stopTimer();
+    timerInterruptHandlers[getTimerInterruptIndex(timerPeripheral)] = irqHandler;
+    initTimer(timerPeripheral, this->clockPeriod);
+}
+
+void Timerf302x8::stopTimer() {
+    HAL_TIM_Base_DeInit(this->halTimer);
```
ddp5730, line 157: Do we want to completely de-init the timer here?  Another function call is `HAL_TIM_Base_Stop_IT()` which just pauses the timer module.  Realistically the functionality appears to be the same either way.

cbolles: We don't have to complete de-init. I imagine this will be used like a "pause" for the timer.

cbolles (changes requested):

ddp5730:
I forgot to implement the one off timer.  Therefore converting to a draft.

src/dev/platform/f3xx/f302x8/Timerf302x8.cpp
ddp5730: I agree.

cbolles (approved):
Looks good!

PR #25, opened 2021-10-22
Feature/aarc10/button

AarC10:
Header file for a button. Was looking for some feedback, suggestions and things I need to fix.

1. Created a button class
2. Read the state of a button
3. Read and write the state of a (possible) LED on the button
4. Debounce the signal from button presses


AarC10:
- Implemented a function that returns a button state
- Changed ActiveState variable to LogicLevelOnPress to be more understandable
- Added a sample to test buttons and the new button interface should be working

include/EVT/dev/button.hpp
```diff
@@ -0,0 +1,59 @@
...
+         * @return The button state.
+         */
+    IO::GPIO::State readButton();
+
+    /**
+         * Time how long the button is in a high state
+         *
+         * @return Time in milliseconds
+         */
+    uint32_t debounce();
```
cbolles, line 43: Are the debounce methods implemented? If not I would look into implementing them using [HAL_GetTick](https://stackoverflow.com/questions/42747128/does-hal-gettick-return-ticks-or-milliseconds-and-how-to-measure-in-microsec) method to get the elapsed time since startup in milliseconds.

include/EVT/dev/buttonLEDs.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define _EVT_buttonLEDs_
+
+#include <EVT/dev/LED.hpp>
+#include <EVT/dev/button.hpp>
+#include <EVT/io/GPIO.hpp>
+#include <stdint.h>
+
+namespace EVT::core::DEV {
+
+class ButtonLEDs : Button {
```
cbolles, line 11: Do we have the ButtonLEDs implemented? If not, we probably shouldn't have the header for it

cbolles (changes requested):
Looks really good. Have some suggested changed, but nice work.

include/EVT/dev/buttonLEDs.hpp
AarC10: ButtonLEDS is most likely unnecessary and will be deleted since there is already an LED module in EVT-Core

include/EVT/dev/button.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+         * @brief Set the Button State object
```
cbolles, line 46: Also, looks like the comments are tabbed over beyond our style

include/EVT/dev/button.hpp
```diff
@@ -0,0 +1,59 @@
...
+         * @param logicLevelOnPress[in] Represents if the button is active high or active low
+         */
+    Button(EVT::core::IO::GPIO& gpio, Button::LogicLevelOnPress logicLevelOnPress);
+
+    /**
+         * Reads the state of the button
+         *
+         * @return The button state.
+         */
+    IO::GPIO::State readButton();
```
ddp5730, line 36: It may be overly pedantic, but I'm not sure readButton() should return a `IO::GPIO::State`.  We are looking for a value of pressed or not pressed here and returning a GPIO value of HIGH or LOW may cause some confusion.

samples/button/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+    DEV::LED led(ledGPIO, DEV::LED::ActiveState::HIGH);
+
+    DEV::Button button = DEV::Button(buttonGPIO, DEV::Button::LogicLevelOnPress::HIGH);
+
+    while (true) {
+        IO::GPIO::State buttonStatus = button.readButton();
+        uart.printf("Button Status: %d \n", buttonStatus);
+
+        led.setState(buttonStatus);
+    }
```
ddp5730, line 27: I would recommend adding at least a wait call here just to slow down the printing a little.  Reading and printing as fast as the uC allows may be excessive.

ddp5730:
Simply reading the button state may not be as useful as triggering events when a button is pressed.  Do we want to add this task to the backlog?

AarC10:
Thanks for the feedback. I have tested the readButton() function with hardware. I have done some basic fixes to the coding style and improved the samples. I will work on the readButton function and I am still in the process of figuring out and creating a debounce function.

AarC10:
Redid the button class.
Currently takes in a GPIO pin and can read the state of the GPIO and has a function for debouncing with a user defined debounce time.

cmake-build-debug/CMakeFiles/3.21.1/CMakeSystem.cmake
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_HOST_SYSTEM "Linux-5.15.16-1-MANJARO")
+set(CMAKE_HOST_SYSTEM_NAME "Linux")
```
cbolles, line 2: Add `cmake-build-debug/` to .gitignore

include/EVT/dev/button.hpp
```diff
@@ -0,0 +1,25 @@
...
+// Forward declarations:
+// The different pins are hardware specific. Forward declaration to allow
+// at compilation time the decision of which pins should be used.
+
+class Button {
+public:
+    Button(IO::GPIO& gpio);
+
+    IO::GPIO::State getState();
+    bool debounce(int debounceTime);
```
cbolles, line 17: Use something like a `uint32_t` here since deboundTime cannot be negative

include/EVT/dev/button.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+private:
+    IO::GPIO& gpio;
+    int timeSinceLastPress;
```
cbolles, line 21: Same here, use a `uint32_t`

src/dev/button.cpp
```diff
@@ -0,0 +1,25 @@
...
+    this->gpio = gpio;
+    this->timeSinceLastPress = 0;
+}
+
+IO::GPIO::State Button::getState() {
+    return this->gpio.readPin();
+}
+
+bool Button::debounce(int debounceTime) {
+    if (HAL_GetTick() - this->timeSinceLastPress > debounceTime) {
```
cbolles, line 15: This is ok for now, but this implementation is specific to the STM (`HAL_GetTick` won't exist in other platforms). We have a task in Trello for a platform independent `GetTick` equivalent. For now we can accept this

cbolles (changes requested):
A couple of changes, it looks like a generated `cmake-build-debug` was added and it looks like the `canopen` submodule was changed when it shouldn't be. Make sure to update your submodules to reflect main.

src/dev/button.cpp
AarC10: Ok

cbolles (approved):
Looks good

mjmagee991 (approved):
Looks great

ddp5730 (approved):
Nice work

PR #27, opened 2021-11-06
<3 EEPROM? <3

mjmagee991:
Created an EEPROM abstract class and concrete implementation of it for the STM M24C32-RMN6P EEPROM
Added memory read/write functionality to the I2C abstract class and I2Cf302x8 concrete class
Added functionality for the EEPROM and M24C32 classes to read and write individuals and arrays of bytes, half words, and words


include/EVT/dev/storage/EEPROM.hpp
```diff
@@ -0,0 +1,119 @@
...
+    virtual uint32_t readWord(uint32_t address) = 0;
+
+    /**
+     * Read a number of consecutive bytes starting at a given memory address.
+     *
+     * @param[in] address The memory address to start reading from
+     * @param[in] numBytes The number of bytes to read
+     * @param[out] buffer Buffer to output bytes to
+     */
+    virtual void readBytes(uint8_t address, uint8_t numBytes, uint8_t* buffer) = 0;
```
cbolles, line 45: NIT: In similar usage (I2C) we have the format with the buffer parameter first then the number of bytes. It's a common practice in most code bases

include/EVT/io/I2C.hpp
```diff
@@ -97,6 +97,48 @@ class I2C {
     void readReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
             uint8_t* bytes, uint8_t length);
 
+    /**
+     * Write a single byte to a register in memory.
```
cbolles, line 101: I may add to the comment a little more clarification (or a link to some explanation) that memory modules follow a slightly different I2C pattern with the reads and writes.

samples/i2c/main.cpp
```diff
@@ -16,11 +16,9 @@ namespace IO = EVT::core::IO;
 namespace time = EVT::core::time;
 
 /** The address of the arduino listening for I2C requests */
-constexpr uint8_t I2C_SLAVE_ADDR = 0x04;
```
cbolles, line 19: Did you intend to modify the I2C sample for this code review?

cbolles (commented):
Looks pretty good, just some suggestions

cbolles (approved):

PR #28, opened 2021-11-08
Linkin Logs

cbolles:
Quick tweak that removes unused code from libraries during compilation, significantly saves space in final binary.


ddp5730 (approved):
I didn't know linkers had garbage collection, but I love it.

PR #30, opened 2021-11-13
LogChamp

mjmagee991:
Implemented Global Logging

Created a Logger class in namespace EVT::core::log
Implemented log levels, timestamps, string formatting, and conditional compilation in the class
Created a global instance of the Logger
Wrote a sample file to show off the Logger's functionality


CMakeLists.txt
```diff
@@ -75,6 +76,7 @@ cmake_dependent_option(EVT_BUILD_SAMPLES
 # TODO: This should be set by the user of this library
 add_compile_definitions(STM32F302x8)
 add_compile_definitions(USE_HAL_DRIVER)
+add_compile_definitions(EVT_CORE_LOG_ENABLE)
```
cbolles, line 79: Ideally this would be a definition set by the user, etc 
```
cmake -DEVT_CORE_LOG_ENABLE ..
```

include/EVT/utils/log.hpp
```diff
@@ -0,0 +1,69 @@
...
+#define EVT_LOG_
+
+#include <cstdint>
+#include <EVT/io/UART.hpp>
+#include <EVT/dev/RTC.hpp>
+
+namespace dev = EVT::core::DEV;
+
+namespace EVT::core::log {
+    class Logger {
```
cbolles, line 11: Minor stylistic thing, we don't have the classes indented inside of the name space, so we do
```
namespace EVT::core::log {

class Logger {
public:
    void log(...);
};

}
```

I would also check the output of cpplint to make sure you are matching the syle guide we follow.

cbolles (commented):
Looks really good, couple of minor suggestions, but otherwise very nice

cbolles (approved):
Looks good!

PR #31, opened 2021-11-16
EEpain

cbolles:
Address issue with writing beyond the starting page


samples/eeprom/main.cpp
```diff
@@ -22,20 +22,28 @@ constexpr uint8_t I2C_SLAVE_ADDR = 0x50;
 constexpr uint8_t BYTE_ADDRESS = 0x00;
 constexpr uint8_t HALF_WORD_ADDRESS = 0x01;
 constexpr uint8_t WORD_ADDRESS = 0x03;
-constexpr uint8_t BYTE_ARR_ADDRESS = 0x07;
```
mjmagee991, line 25: I don't know if this is excessive, but I was thinking of having at least one example to show off every part of this feature. That's why I made examples for each of the methods being used. I think it would make more sense to leave this byte array example in and have a separate example or two to show writing over a page break. Conceptually, they seem like different parts of the feature, so it seems to me that they should be demonstrated with two separate examples to avoid confusion. To do this, you could put the example I already had and rename your example something like PAGE_BREAK_EXAMPLE_ADDRESS.

src/dev/storage/platform/M24C32.cpp
```diff
@@ -25,21 +25,36 @@ uint32_t M24C32::readWord(uint32_t address) {
 }
```
mjmagee991, line 25: writeHalfWord(), readHalfWord(), writeWord(), and readWord() all have the potential to cause the same problem as reading/writing the arrays did, so they should call readBytes() instead of i2c.readMemReg().

mjmagee991 (changes requested):
Looks pretty good. There are just a couple details I think could be improved.

samples/eeprom/main.cpp
cbolles: Yea that's a good point, those two features are now broken into two different tests

src/dev/storage/platform/M24C32.cpp
cbolles: Good catch! Those now use the read/write bytes

ddp5730 (approved):

zsteig (approved):

mjmagee991 (approved):

PR #32, opened 2021-12-19
Clang for Matt

mjmagee991:
Added clang-format linting to EVT-core
Created a .clang-format file to store formatting settings
Added CMake integration of clang-format to apply formatting upon building
Applied formatting to all files in EVT-core


cmake/evt-core_clang-format.cmake
```diff
@@ -0,0 +1,46 @@
...
+# Get all project files
+file(GLOB_RECURSE
+        ALL_CXX_SOURCE_FILES
+        *.[chi]pp *.[chi]xx *.cc *.hh *.ii *.[CHI]
+        )
+
+# Get the path to clang-format
+find_program(CLANG_FORMAT "clang-format")
+# Run the formatting if clang-format is found
+if (CLANG_FORMAT)
```
cbolles, line 17: After doing a bit more digging, I think the more "cmake like" approach would be to add a custom target seen [here](https://cmake.org/cmake/help/latest/command/add_custom_target.html). [This is an example](https://gist.github.com/baiwfg2/39881ba703e9c74e95366ed422641609) of adding custom commadns

cbolles (changes requested):
Looks really good, my bad for changing things up a bit, but I think we should look into making the formatting a cmake target.

cmake/evt-core_clang-format.cmake
mjmagee991: I changed things around to work as a custom target with custom commands instead of using execute_process. I had to add a weird workaround because Windows has a character limit for its command line, but it works on both Linux and Windows now.

cbolles (approved):
This looks great!

PR #36, opened 2022-01-26
Bug fix/dpopp/adc dma

ddp5730:
Fixed the bug where the ADC would not generate the correct values.  
- The main fix applied was to only run the ADC for the number of conversions as ADC channels configured.  
- The DMA buffer was also updated to have this same effective size.
- Some modifications were also made to stop and de-init peripherals when adding new channels.

### DIFF
--- a/include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
+++ b/include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
@@ -43,9 +43,9 @@ public:
 private:
     // Max number of channels supported by the ADC
     static constexpr uint8_t MAX_CHANNELS = 15;
-    // Max voltage of the ADC
-    static constexpr float MAX_VOLTAGE = 3.6;
-    // Max value for a 12 bit ADC reading
+    // Positive reference voltage of the ADC.  Needs to be updated based on the hardware configuration
+    static constexpr float VREF_POS = 3.3;
+    // Max value for a 12 bit ADC reading (2^12 - 1)
     static constexpr uint32_t MAX_RAW = 4095;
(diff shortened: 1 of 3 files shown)


include/EVT/io/platform/f3xx/f302x8/ADCf302x8.hpp
```diff
@@ -43,9 +43,9 @@ class ADCf302x8 : public ADC {
 private:
     // Max number of channels supported by the ADC
     static constexpr uint8_t MAX_CHANNELS = 15;
-    // Max voltage of the ADC
-    static constexpr float MAX_VOLTAGE = 3.6;
-    // Max value for a 12 bit ADC reading
+    // Positive reference voltage of the ADC.  Needs to be updated based on the hardware configuration
+    static constexpr float VREF_POS = 3.3;
```
ddp5730, line 47: So the voltage reading is dependent on the VREF+ pin for the ADC.  As best as I can tell this is the VDDA uC pin (1.8V < VDDA < 3.6V).  Essentially this is the positive uC supply voltage, but the uC doesn't actually know this value.  

I'm open suggestions on how to handle this.

cbolles: I think later a feature can be opened to include setting the VREF value

cbolles: [Trello Card](https://trello.com/c/wyGLJDR0) made for this

cbolles (approved):
Looks great! Great work

mjmagee991 (approved):
Looks good!

PR #38, opened 2022-01-29
Feature/mjh9585/spi

mjh9585:
Added hardware drivers for Full-duplex and half-duplex SPI. Methods added:
### Constructor
- SPI(GPIO **CSPins, uint8_t pinLength, Pin sckPin, Pin mosiPin, Pin misoPin); //For Full-duplex SPI
- SPI(GPIO **CSPins, uint8_t pinLength, Pin sckPin, Pin mosiPin); //For Half-duplex SPI
### Low Level
- startTransmission(uint8_t device); // starts the SPI communication with a device. Must be called before any low level calls.
- endTransmission(uint8_t device); // ends the SPI communication with a device. Must be called a after an SPI communication.
- write(uint8_t byte); // writes a single byte to a device.
- write(uint8_t *bytes, uint8_t length); // write multiple bytes to a device.
- read() // read a single byte to a device.
- read(uint8_t *bytes, uint8_t length); // read multiple bytes to a device.
### Higher level
- configureSPI(uint32_t baudRate, uint8_t mode, uint8_t order); //configures the transmission settings for the SPI.
- writeReg(uint8_t device, uint8_t reg, uint8_t byte); // writes a byte to a device register.
- writeReg(uint8_t device, uint8_t reg, uint8_t *bytes, uint8_t length); // writes multiple bytes to a device starting at a register.
- readReg(uint8_t device, uint8_t reg); // reads a byte from a device register.
- readReg(uint8_t device, uint8_t reg, uint8_t *bytes, uint8_t length); // reads multiple bytes from a device starting at a register.
### Sample Code
Two samples are included from testing: 
- spi
    - simple test of all of the SPI functions between the Nucleo and an Arduino.
- spi_ADXL345
    - small test of the low level methods to communicate with an ADXL345 accelerometer.


src/io/platform/f3xx/f302x8/SPIf302x8.cpp
```diff
@@ -0,0 +1,359 @@
...
+     * @param pinLength the number of pins in the chip select array
+     * @param sckPin the pin for the clk line
+     * @param mosiPin the mosi pin for sending data
+     * @param misoPin the miso pin for receiving data
+     */
+SPIf302x8::SPIf302x8(GPIO** CSPins, uint8_t pinLength, Pin sckPin, Pin mosiPin, Pin misoPin) : SPI(CSPins, pinLength, sckPin, mosiPin, misoPin) {
+    uint8_t mosiPort = getMOSIPortID(mosiPin);
+    uint8_t misoPort = getMISOPortID(misoPin);
+    uint8_t sckPort = getSCKPortID(sckPin);
+    GPIO_InitTypeDef GPIOInit = {0};
```
cbolles, line 75: For the GPIO setup. We now have have a method for abstracting the GPIO init logic. See [this function](https://github.com/RIT-EVT/EVT-core/blob/main/include/EVT/io/platform/f3xx/f302x8/GPIOf302x8.hpp#L69)

include/EVT/io/SPI.hpp
```diff
@@ -0,0 +1,143 @@
...
+    /** The SPI clock line */
+    Pin sckPin;
+    /** The MOSI data line*/
+    Pin mosiPin;
+    /** The MISO data line */
+    Pin misoPin;
+
+protected:
+    uint8_t CSPinsLength;
+    GPIO* CSPins;
```
cbolles, line 138: Should this be `GPIO* CSPins[]`?

include/EVT/io/SPI.hpp
```diff
@@ -0,0 +1,143 @@
...
+public:
+    /**
+         * Make an instance of an I2C interface that will use the given pins
+         * for clock and data lines.
+         *
+         * @param sckPin[in] The SPI clock pin
+         * @param mosiPin[in] The mosi pin (data out)
+         * @param misoPin[in] The miso pin (data in)[optional]
+         */
+    SPI(GPIO** CSPins, uint8_t pinLength, Pin sckPin, Pin mosiPin, Pin misoPin);
```
cbolles, line 40: Somethings to note.

Having the CSPins passed in and set via this manner will require the array of device pins to remain valid memory for the life time of the SPI device. An example where this might not be the case is if the SPI device is created within a function with the device pins stored as a local variable.

In general, you'll see in embedded development, the parameter will be an array of device pins, and those pins will then be copied into an array of device pins that is stored within the SPI device. That way, you know the device pins will live as long as the SPI interface. That does mean we will also have an upper bound on how many devices will be selected. But below is some example psuedo code

```
SPI(GPIO CSPins[], uint8_t numPins) {
    for(int i = 0; i < numPins; i++) {
        this->pins[i] = CSpins[i];
    }
}
```
Where this->pins would be an instance variable with a maximum size.

Another thing that should be considered is instead of having a list of pins as an input. The SPI device could take a `Mux` object as an input. The idea being that the `Mux` could be implemented in different ways (via inheritance). That way if say the chip select is actually done through a GPIO expander (common) or demux (very common), the SPI device is immune to those changes. That being said, that can be a feature/modification added later.

Also this was a huge wall of text, so feel free to ask for clarification.

include/EVT/io/SPI.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+    SPI(GPIO** CSPins, uint8_t pinLength, Pin sckPin, Pin mosiPin);
+
+    /**
+         * begin a device transmission. Call before each set of read and write interactions.
```
cbolles, line 45: Note: clang-format doesn't format comments, so that is why the comments now looked over tabbed. Unfortunately we do have to change that manually.

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/**
+     * writes an array of bytes to the SPI device
+     * @param device the device to write to in CSPins
+     * @param bytes an array of bytes of length n to write to SPI device
+     * @param length the length of the array
+     */
+void SPIf302x8::write(uint8_t* bytes, uint8_t length) {
+    //if (startTransmission(device)) {
```
cbolles, line 341: If you don't need the code anymore, remember to delete commented out code (or if it's important to have the commented code for later implementation, add comments explaining it.

cbolles (changes requested):
Overall looks really good. Added some suggestions, but mainly the ones below are the important ones.

* Look into copying the CSpins into an instance variable
* Use the new method for GPIO initialization

ddp5730 (changes requested):
I agree with Collin's comments.

include/EVT/io/SPI.hpp
mjh9585: Sounds good. Since we aren't using much SPI right now, would 5 be a good maximum size for now or should we start larger?

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
mjh9585: Just missed that in my cleaning up!

include/EVT/io/SPI.hpp
mjh9585: yeah, I wasn't expecting the code to get reformatted in the pull request.

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
mjh9585: Will this function take into account the specific pins that have SPI hardware and that all the pins are on the same SPI port?

include/EVT/io/SPI.hpp
cbolles: Yeah I think 5 is reasonable, you can have it in a `constexpr` so it's easy to update later

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
cbolles: It doesn't it really replaces lines 96-129.

include/EVT/io/SPI.hpp
mjh9585: cool

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
mjh9585: oh, ok

cbolles (approved):
Looks really good!

include/EVT/io/SPI.hpp
```diff
@@ -0,0 +1,151 @@
+#ifndef _EVT_SPI_
+#define _EVT_SPI_
+
+#include "GPIO.hpp"
```
mjmagee991, line 4: NIT: This should likely be an include with <> notation.

include/EVT/io/SPI.hpp
```diff
@@ -0,0 +1,151 @@
...
+     * Constructs an SPI instance in half duplex mode to only send data
+     * @param CSPins an array of chip select pins for selecting which device to communicate with.
+     * @param pinLength the number of pins in the chip select array
+     * @param sckPin the pin for the clk line
+     * @param mosiPin the mosi pin for sending data
+     */
+    SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin);
+
+    /**
+     * begin a device transmission. Call before each set of read and write interactions.
```
mjmagee991, line 52: NIT: The block comments are normally capitalized. It's probably fine if you leave them lowercase if you make them consistent. There's also normally an empty line between the explanation of the method and the parameters/return section of the comment.

src/io/SPI.cpp
```diff
@@ -0,0 +1,74 @@
+//
+// Created by matth on 10/2/2021.
```
mjmagee991, line 2: NIT: We normally don't have a comment at the beginning of cpp files.

src/io/platform/f3xx/f302x8/SPIf302x8.cpp
```diff
@@ -0,0 +1,313 @@
+//
+// Created by Matthew H on 10/2/2021.
```
mjmagee991, line 2: NIT: We normally don't have a comment at the beginning of cpp files.

mjmagee991 (changes requested):
Overall, it looks really good. I just had a couple nitpicks with documentation and one include.

mjmagee991 (approved):
Nice work

PR #39, opened 2022-02-01
I2C and Error

cbolles:
This PR adds error handling to the I2C interface. Each I2C method now returns an I2CStatus. "Return" values are now passed in via a pointer and updated indirectly. This is a common practice in C/C++ especially in the Linux kernel. This allows us to now tell when an error has taken place and respond accordingly.

For now, only the I2C interface has been updated to have the error handling. But in the future all interfaces will have it. So be critical as this system will then be used throughout EVT-core with each interface having a "status" that is returned. This will then extend to include the devices as well.

### DIFF
--- a/include/EVT/io/I2C.hpp
+++ b/include/EVT/io/I2C.hpp
@@ -3,6 +3,13 @@
 
 #include <stdint.h>
 
+#define I2C_RETURN_IF_ERR(func) {\
+    I2C::I2CStatus status = func;\
+    if (status != I2C::I2CStatus::OK) {\
+        return status;\
+    }\
+}
+
 namespace EVT::core::IO {
 // Forward declarations:
 // The different pins are hardware specific. Forward declaration to allow
@@ -11,6 +18,18 @@ enum class Pin;
 
 class I2C {
 public:
+    /**
+     * Represents potential errors that may take place when using the I2C
+     * interface. Each method that interfaces over I2C could potentially
+     * return one of these errors, or OK if no error.
+     */
+    enum class I2CStatus {
+        TIMEOUT = 0,
+        BUSY = 1,
+        ERROR = 2,
+        OK = 3
+    };
+
     /**
      * Make an instance of an I2C interface that will use the given pins
      * for clock and data lines.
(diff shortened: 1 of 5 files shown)


include/EVT/io/I2C.hpp
```diff
@@ -3,6 +3,14 @@
...
 #include <stdint.h>
 
+#define I2C_RETURN_IF_ERR(func)             \
+    {                                       \
+        I2C::I2CStatus status = func;       \
+        if (status != I2C::I2CStatus::OK) { \
+            return status;                  \
+        }                                   \
+    }
+
```
ddp5730, line 13: Is there any particular reason you opted for a macro here vs. a traditional function definition?

ddp5730 (approved):
Looks great!

include/EVT/io/I2C.hpp
cbolles: Yeah, so basically I want some functions to return a value on error. So if I had this as a function, the function will return the status to the caller, but I want the caller itself to return the status on error. 

```
I2C::I2CStatus myFunc() {
    ...
    I2C_RETURN_IF_ERR(read(...));
    ...
}
```

As a macro, `myFunc()` will return the status if the read fails. Basically it leads to cleaner looking code so I don't have to constantly wrap the I2C calls in if statements.

Additionally, since its a small snippet, the argument could be made about function overhead blah blah blah, bunch of stuff we usually don't worry too much about. Really it's more about what I have above

ddp5730: Ah, that makes perfect sense.

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
```diff
@@ -91,57 +91,99 @@ I2Cf302x8::I2Cf302x8(Pin sclPin, Pin sdaPin) : I2C(sclPin, sdaPin) {
...
 
-void I2Cf302x8::writeMemReg(uint8_t addr, uint32_t memAddress, uint8_t bytes[], uint8_t size, uint16_t memAddSize,
-                            uint8_t maxWriteTime) {
+I2C::I2CStatus I2Cf302x8::writeMemReg(uint8_t addr, uint32_t memAddress,
+                                      uint8_t* bytes, uint8_t size,
+                                      uint16_t memAddSize,
+                                      uint8_t maxWriteTime) {
     uint16_t memAddress16 = memAddress;
-    HAL_I2C_Mem_Write(&halI2C, addr << 1, memAddress16, memAddSize, bytes, size, DEFAULT_I2C_TIMEOUT);
-    HAL_Delay(maxWriteTime);
```
mjmagee991, line 139: This delay is required to give the signal time to complete without throwing errors. I think there should be a way to check flags to do this instead of just a flat delay, but I figured this would be good enough for now. You can either try to figure out a better way of doing it or add the HAL Delay back in.

mjmagee991 (changes requested):
Looks really good. Seems like a great step toward error handling. I just had one concern with the EEPROM write method.

src/io/platform/f3xx/f302x8/I2Cf302x8.cpp
cbolles: Ahh, I had thought the delay was related to the timeout. For waiting for the signal to propagate through, that is usually handled higher up because different systems may require different amounts of delay. (For example the EEPROM responds much faster over I2C then the BQ chip on the BMS). I'll certain re-run tests with the EEPROM without the delay written here to make sure everything is still ok.

cbolles: For now I reverted the change, but we should move any wait time to the EEPROM device since its specific to the EEPROM and not I2C.

mjmagee991 (approved):
Looks good

zsteig (approved):

PR #40, opened 2022-02-01
Real time clock

amenkes319:
A real-time clock that goes off after the clock period is hit. Tests in main contain stoping and starting the clock again and a continuous loop of the timer going off, then resetting it to go off again.


include/EVT/dev/RTCTimer.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+private:
+    /** Instance of on-board*/
+    RTC* m_rtc;
+
+    /**
+     * The amount of seconds that have elapsed while the timer is running.
+     * Only updates when stopTimer() is called.
+     */
+    uint32_t m_time;
```
cbolles, line 82: More of a convention thing, we don't use the `m_` prefix

cbolles (changes requested):
Overall, looks really good, just a small convention thing

include/EVT/dev/RTCTimer.hpp
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @param rtc
+     */
+    RTCTimer(RTC& r);
+
+    /**
+     * Create instance of RTCTimer.
+     *
+     * @param rtc Instance of on-board
+     * @param clockPeriod Amount of time it takes for the time to go off in ms
```
cbolles, line 26: NITPICK: The parameter name is `clock` in the code and `clockPeriod` in the docstring

include/EVT/dev/RTCTimer.hpp
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Gets whether the timer has gone off
+     *
+     * @return whether the time has gone off
+     */
+    bool hasGoneOff();
+
+private:
+    /** Instance of on-board*/
+    RTC* rtc;
```
cbolles, line 76: Instead of using a pointer, I'd recommend using a reference.

cbolles (changes requested):
A couple of small things I just noticed. The NITPICK isn't super important. But in general we use references as much as possible to reduce the change of a null-ptr error.

src/dev/RTCTimer.cpp
```diff
@@ -0,0 +1,46 @@
+#include "EVT/dev/RTCTimer.hpp"
+
+namespace EVT::core::DEV {
+
+RTCTimer::RTCTimer(RTC& r) : rtc(r) {
+    rtc = r;
```
cbolles, line 6: NITPICK: Line 6 is already handled via the `rtc(c)` on line f

cbolles (approved):
Looks good, a couple of nitpicks left, but nothing critical

mjmagee991 (approved):
Looks good

PR #41, opened 2022-02-04
Feature/cbolles/thermistor

cbolles:
Super basic implementation of a thermistor that allows the user to provide the conversion function as a parameter. This should allow flexibility for more complex thermistor conversion functions.

Note. This was implemented using fixed point arithmetic to reduce the overhead added by floats. In the future, once the ADC has the functionality to measure in fixed point the voltage (ADC returns voltage in millivolts rather then as a float) then this can be updated accordingly.

Another note, there seemed to be a bit of a strange bug with the ADC that caused the code to halt (not fault, just stop running main) if there is only one channel. This is why the test has two ADC channels while it only needs 1.


samples/thermistor/main.cpp
```diff
@@ -12,25 +12,40 @@ namespace IO = EVT::core::IO;
 namespace DEV = EVT::core::DEV;
 namespace time = EVT::core::time;
 
+
+/**
+ * Note, this conversion function is just to show off how a conversion
+ * function will loook like. In actuality, this will be determined through
```
AarC10, line 17: Just a minor typo here

AarC10:
Implementation looks good to me

ddp5730:
The ADC bug you described should be resolved here: https://github.com/RIT-EVT/EVT-core/pull/42

samples/thermistor/main.cpp
cbolles: Good catch

ddp5730 (approved):
Looks great.  It's up to you if you want to merge this or wait for #42 first so you can remove that redundant ADC initialization.

samples/thermistor/main.cpp
```diff
@@ -0,0 +1,50 @@
+/**
+ * This example prints out the time as read in from the RTC
+ * every second.
```
mjmagee991, line 3: This comment should be updated.

mjmagee991 (changes requested):
Looks great. I just caught a documentation mistake.

samples/thermistor/main.cpp
cbolles: Good catch

mjmagee991 (approved):
Looks great

PR #42, opened 2022-02-04
Resolved bug where single channel ADC would use all processor resources.

ddp5730:
- Bug was caused by handling the ADC interrupt during the DMA interrupt handler.
- This was added based on comment in `stm32f3xx_hal_adc.c`
   - `(++) Insert the ADC interruption handler function HAL_ADC_IRQHandler()
              into the function of corresponding DMA interruption vector
              DMAx_Channelx_IRQHandler().`


cbolles (approved):

mjmagee991 (approved):
Nice work

ddp5730:
It appears I was too hasty in fixing this bug.  While this solution was valid, the real issue is the ADC is triggering far too many interrupts and I think it should be slowed down.  I'm closing this PR and I'm going to do more thorough analysis before posting what I believe is the more comprehensive fix.

PR #43, opened 2022-02-04
Bug fix/dpopp/adc single channel

ddp5730:
- Configured ADC to run with AHB clock at (8MHz) to slow it down to same clock speed as SYSCLK.
- ADC was previously configured with PLL clock that ran at 32 MHz and generated too many interrupts for the system to handle
   - ADC generates interrupts every conversion and half conversion.


cbolles (approved):
Looks good!

mjmagee991 (approved):
Looks great.

PR #44, opened 2022-02-05
Who took a byte out of my EEPROM size?

cbolles:
Change the size of the address size for reading and writing data from `uint8_t` to `uint32_t` to allow for EEPROM sizes larger then 256 bytes.


mjmagee991 (approved):
I probably should've seen that on the first go around. Good find.

ddp5730 (approved):

PR #45, opened 2022-02-05
Kicking the bigger CAN

cbolles:
Increase the size of CANopen related data. This includes increasing the size of the CANopen buffer and updating the canopen submodule to include code that increases the size of the SDO buffer.


mjmagee991 (approved):
Looks good

zsteig (approved):

PR #46, opened 2022-02-07
Do you have your CAN Bus pass?

zsteig:
Added in functionality to set a filter Id to avoid overwhelming the CAN bus devices. Currently it only allows the setting of a single identifier, however this can be expanded later on if needed. Messages with the CANopen emergency code are always allowed through.


src/io/platform/f3xx/f302x8/CANf302x8.cpp
```diff
@@ -101,19 +101,31 @@ CANf302x8::CANf302x8(Pin txPin, Pin rxPin, bool loopbackEnabled)
...
-    defaultFilter.FilterMaskIdHigh = 0;
-    defaultFilter.FilterMaskIdLow = 0;
-    defaultFilter.FilterFIFOAssignment = CAN_FILTER_FIFO0;
-    defaultFilter.FilterBank = 0;
-    defaultFilter.FilterMode = CAN_FILTERMODE_IDMASK;
-    defaultFilter.FilterScale = CAN_FILTERSCALE_32BIT;
-    defaultFilter.FilterActivation = ENABLE;
-
-    HAL_CAN_ConfigFilter(&halCAN, &defaultFilter);
+    CAN_FilterTypeDef emergencyFilter;
```
cbolles, line 104: For these filters, the user of the code should be able to pass these in more explicitly. Filtering based on ID is good for the case of getting the messages targeting a specific device, but we also want the ability to letting in messages coming from another device. Ultimately what we would want is something along the lines of

```
can.addExplicitFilter(0x05); // Will let in messages with the CAN ID 0x05
can.addExplicitFilter(0x06); // Will let in message with the CAN ID 0x06
can.addMaskFilter(0x83); // Will add this mask as a filter
```

That way we can say explicit allow for the TMS to get messages that correspond to BMS temperature data. Additionally this will apply better in places were we use CAN, but not necessarily CANopen (like working with the ground fault detection board).

src/io/platform/f3xx/f302x8/CANf302x8.cpp
```diff
@@ -168,6 +180,10 @@ CANMessage* CANf302x8::receive(CANMessage* message, bool blocking) {
     }
 }
 
+void CANf302x8::setCANFilterId(uint32_t identifier) {
+    this->identifier = identifier;
```
cbolles, line 184: This identifier would not be explicitly needed with the proposed method for adding filters.

cbolles (changes requested):
Good start, but we will want to more generalize the filter logic.

include/EVT/io/platform/f3xx/f302x8/CANf302x8.hpp
```diff
@@ -44,6 +44,16 @@ class CANf302x8 : public CAN {
      */
     CANMessage* receive(CANMessage* message, bool blocking = false);
 
+    /**
+     * @copydoc EVT::core::IO::CAN::addCANFilter
```
cbolles, line 48: The copy docs have been removed since the docstrings are automatically inherited, this should already be the case in `main` so you may just need to merge.

cbolles (changes requested):

zsteig:
On my local machine I get no errors when running `make -j`

I did get the `undefined reference to `vtable for EVT::core::IO::CAN'` but I was able to resolve it by deleting and remaking the build folder.

So I am not sure why this error is appearing in the GitHub pipeline

cbolles (approved):

mjmagee991 (approved):
Looks good

PR #50, opened 2022-02-18
No IT for this (CAN) Network

cbolles:
Adds error handling for the CAN interface.

* Creates an enum for possible CAN errors
* Have each method return a CAN status
*  **New methods connect and disconnect** so that we can handle possible HAL errors starting CAN
* Update samples to handle possible errors

### DIFF
--- a/include/EVT/io/CAN.hpp
+++ b/include/EVT/io/CAN.hpp
@@ -26,6 +26,16 @@ enum class Pin;
 class CAN {
 public:
     /**
+     * Represents potential errors that may take place when using the CAN
+     * interface.
+     */
+    enum class CANStatus {
+        TIMEOUT = 0,
+        ERROR = 1,
+        OK = 3
+    };
+
+    /**
      * Creates a new instance of the CAN interface which will use the given
      * transmit and receive pins.
      *
@@ -39,8 +49,9 @@ public:
      * Transmit the message over CAN.
      *
      * @param[in] message The message to send over CAN.
+     * @return The status associated with sending the message
      */
-    virtual void transmit(CANMessage& message) = 0;
+    virtual CANStatus transmit(CANMessage& message) = 0;
 
     /**
      * Receive a message over CAN. The user can either receive in blocking or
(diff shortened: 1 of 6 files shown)


include/EVT/io/CAN.hpp
```diff
@@ -25,6 +25,16 @@ enum class Pin;
  */
 class CAN {
 public:
+    /**
+     * Represents potential errors that may take place when using the CAN
+     * interface.
+     */
+    enum class CANStatus {
+        TIMEOUT = 0,
```
mjmagee991, line 33: Is there any particular reason you chose these values for the enum? Isn't 0 the standard value for a successful status? In most cases, this shouldn't matter, but if you set okay to 0, you could use the enum as a boolean to have a simpler if statement to check for success (I think). I'm not certain C++ would do the implicit casting to check if the enum value is zero or nonzero to check truth or falsity, so this might be irrelevant.

include/EVT/io/platform/f3xx/CANf3xx.hpp
```diff
@@ -34,9 +34,29 @@ class CANf3xx : public CAN {
      */
     CANf3xx(Pin txPin, Pin rxPin, bool loopbackEnabled = false);
 
-    void transmit(CANMessage& message);
+    /**
+     * Attempt to join the CAN network.
+     *
+     * For the STM32f3xx this involves attempting to startup the CAN
+     * interface. This chould cause an error in the case of invalid
```
mjmagee991, line 41: This chould either be could or should.

mjmagee991 (changes requested):
Looks really good. The error handling should definitely come in handy. I just had two nitpicks.

include/EVT/io/CAN.hpp
cbolles: That's a good point, I'll change the numeric representation

include/EVT/io/platform/f3xx/CANf3xx.hpp
cbolles: Lol good catch

ddp5730 (approved):

PR #51, opened 2022-02-18
Botulism

cbolles:
Handle issue with the FixedQueue that caused a HardFault. Improper checking of the end of queue marker resulting in bad memory reads. Tested by having the loopback run for 20 minutes without hardfault.


mjmagee991 (approved):
Botulism

zsteig (approved):

PR #52, opened 2022-02-19
CMake up your mind

cbolles:
Move the logic for selecting the board into the compiler cmake file so that other projects can select different platforms to build for.


AarC10 (approved):
Looks fine to me. Good job.

mjmagee991 (approved):
Nice work

PR #53, opened 2022-02-26
Millis

cbolles:
Adds a function to EVT-core for getting the current system time in milliseconds. This is just a wrapper around `HAL_GetTick` for exposing a common interface.


zsteig (approved):

samples/millis/main.cpp
```diff
@@ -0,0 +1,38 @@
+/**
```
mjmagee991, line 1: This file needs an update to its documentation.

mjmagee991 (changes requested):
Looks good. Just seems like you missed the documentation in the sample file.

samples/millis/main.cpp
cbolles: Nice catch

mjmagee991 (approved):
Nice work

PR #54, opened 2022-03-15
Pushing 🅿️WM Improvements

AarC10:
- Optimized the PWM module by replacing float usages with uint32_t
- Microseconds are passed in to the setPeriod function now
- Duty Cycle percentage is now specified as an integer

### DIFF
--- a/include/EVT/io/PWM.hpp
+++ b/include/EVT/io/PWM.hpp
@@ -23,14 +23,14 @@ public:
     /**
      * Set the duty cycle for the pin to operate at.
      *
-     * @param[in] dutyCycle Duty cycle to set the pin to.
+     * @param[in] dutyCycle Duty cycle as a whole number to set the pin to.
      */
     virtual void setDutyCycle(uint32_t dutyCycle) = 0;
 
     /**
-     * Set the period for the PWM in seconds.
+     * Set the period for the PWM in microseconds.
      *
-     * @param[in] period The period of the PWM in seconds.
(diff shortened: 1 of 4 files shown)


include/EVT/io/PWM.hpp
```diff
@@ -25,21 +25,21 @@ class PWM {
      *
      * @param[in] dutyCycle Duty cycle to set the pin to.
      */
-    virtual void setDutyCycle(float dutyCycle) = 0;
+    virtual void setDutyCycle(uint32_t dutyCycle) = 0;
 
     /**
      * Set the period for the PWM in seconds.
```
cbolles, line 31: Will probably need to change this comment to reflect the new units. True for other changed methods as well

src/io/platform/f3xx/PWMf3xx.cpp
```diff
@@ -248,7 +248,7 @@ void PWMf3xx::setPeriod(float period) {
     // autoreload value into a valid range.
     do {
         prescaler++;
-        autoReload = period * clockFrequency / (prescaler + 1) - 1;
+        autoReload = clockFrequency / (prescaler + 1) - 1 / period;
```
cbolles, line 251: Now that we divide by the period here, we probably want to handle the case where someone enters 0 as a period

cbolles (changes requested):
Looks good! Just a couple of changes

cbolles (approved):
Looks good!

mjmagee991 (approved):
Looks good

PR #55, opened 2022-03-15
Can I copy your button debounce? Sure, just change it so it isn't obvious. 

AarC10:
- Converted HAL_GetTick calls used in the button class and its samples to use EVT's millis function found in time utils

Screenshot_20220315_180331


mjmagee991 (approved):
Nice

cbolles (approved):
Great job! Nice catching removing the include as well

PR #56, opened 2022-03-19
Polyamorous Building

mjmagee991:
Added GitHub build testing for the F334


cbolles (approved):
Looks good!

zsteig (approved):

cbolles (changes requested):
`-B` just specifies the build path, unfortunately it doesn't actually run the build. We might be able to do something like
```
mkdir build && cd build && cmake ../ -DTARGET_DEV=STM32F302x8 && make -j
```

cbolles (approved):
Looks good

PR #57, opened 2022-03-19
mjh9585/f334 SPI

mjh9585:
Added SPI support for the STM32F334x8. Also added the template for SPI in Half-Duplex mode.

### DIFF
--- a/samples/spi/ADXL345/main.cpp
+++ b/samples/spi/ADXL345/main.cpp
@@ -26,26 +26,15 @@ IO::GPIO* devices[deviceCount];
 int main() {
     // Initialize system
     IO::init();
-#ifdef STM32F302x8
-    //CS: D5
-    devices[0] = &IO::getGPIO<IO::Pin::PB_4>(EVT::core::IO::GPIO::Direction::OUTPUT);
+
+    //CS: D10
+    devices[0] = &IO::getGPIO<IO::Pin::SPI_CS>(EVT::core::IO::GPIO::Direction::OUTPUT);
     devices[0]->writePin(EVT::core::IO::GPIO::State::HIGH);
 
-    //CLK: D3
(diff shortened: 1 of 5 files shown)


samples/spi/ADXL345/main.cpp
```diff
@@ -34,6 +35,17 @@ int main() {
     //MOSI: CN7-3
     //MISO: CN7-2
     IO::SPI& spi = IO::getSPI<IO::Pin::PB_3, EVT::core::IO::Pin::PC_12, EVT::core::IO::Pin::PC_11>(devices, deviceCount);
+#endif
+#ifdef STM32F334x8
+    //CS: D7
+    devices[0] = &IO::getGPIO<IO::Pin::PB_10>(EVT::core::IO::GPIO::Direction::OUTPUT);
```
cbolles, line 41: To avoid conditional compilation here, you can also add the "default" SPI pins in `pin.hpp` which currently has conditional compilation for 334 vs 302 pins. That would make the samples more simple which is usually the goal

cbolles (changes requested):
Looks good! One small suggestion to reduce complexity in the samples

samples/spi/ADXL345/main.cpp
mjh9585: Sounds good, are there any specific SPI pins that should be considered the "default"?

cbolles: I'd recommend using the pins that are labeled on the "Arduino" compatible portion of the STM headers, but really it can be whatever you decide.

mjh9585: ok

mjmagee991 (approved):
Looks good

cbolles (approved):

PR #58, opened 2022-03-21
"I2C doesn't work", "Well did you turn it on?"

cbolles:
Enables the clock for the I2C in the 334.

NOTE: The system won't compile for the 334 until the SPI PR is merged in.


mjmagee991 (approved):
Nice catch

zsteig (approved):

PR #59, opened 2022-03-27
Fixed Fixed Point Point

mjmagee991:
Fixed the PWM equations to work with fixed point correctly

### DIFF
--- a/samples/pwm/main.cpp
+++ b/samples/pwm/main.cpp
@@ -14,10 +14,36 @@ int main() {
     IO::init();
 
     IO::PWM& pwm = IO::getPWM<IO::Pin::PC_0>();
+    IO::UART& uart = IO::getUART<IO::Pin::UART_TX, IO::Pin::UART_RX>(9600);
     // 1000000 microseconds (1 second) period
-    pwm.setPeriod(1000000);
+    pwm.setPeriod(100000);
     // 50 % duty cycle
     pwm.setDutyCycle(50);
+
+    uint32_t autoReload = 0;
+    uint32_t prescaler = -1;
+    uint32_t clockFrequency = HAL_RCC_GetSysClockFreq();
+
+    // Required loop in order to determine a prescaler which will bring the
+    // autoreload value into a valid range.
+    do {
+        prescaler++;
+        autoReload = (uint32_t) (((uint64_t) (clockFrequency / (prescaler + 1))) * 1000000 / pwm.getPeriod());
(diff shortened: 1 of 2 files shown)


AarC10 (approved):

AarC10:
Seems good to me. May I ask what was the issue with the autoreload?

mjmagee991:
> Seems good to me. May I ask what was the issue with the autoreload?

Before you changed it, the equation was `autoReload = period * clockFrequency / (prescaler + 1) - 1;`
After you changed it, the equation was `autoReload = clockFrequency / (prescaler + 1) - 1 / period;`

This is still algebraically correct, but C++ applies integer division to any expression where two integers are involved, which basically means it always rounds to an integer. Because period is always > 0, 1 / period will always evaluate to either 1 or 0, which was the problem. 

It also seems that you didn't account for the fact that period is now in microseconds instead of seconds, so you have to cancel that out by dividing the first term by 1000000 (microseconds/second).

The solution I came up with was `autoReload = period * (uint64_t) clockFrequency / (prescaler + 1) / 1000000 - 1;`
This takes the value `period * clockFrequency / (prescaler + 1)` from the original equation and divides by 10^6 to counteract the  fact that it is now in microseconds. I also cast to `uint64_t` because `period * clockFrequency` could overflow 32 bits when `period` is in microseconds, which would cause a problem.

AarC10:
Alright thanks for the explanation. Great work!

zsteig (approved):

samples/pwm/main.cpp
```diff
@@ -14,11 +14,37 @@ int main() {
...
 
     IO::PWM& pwm = IO::getPWM<IO::Pin::PC_0>();
+    IO::UART& uart = IO::getUART<IO::Pin::UART_TX, IO::Pin::UART_RX>(9600);
     // 1000000 microseconds (1 second) period
-    pwm.setPeriod(1000000);
+    pwm.setPeriod(100000);
     // 50 % duty cycle
     pwm.setDutyCycle(50);
 
+    uint32_t autoReload = 0;
```
cbolles, line 23: You probably want to remove this test code, or if you wanted to keep it, move it to its own test.

cbolles (changes requested):

samples/pwm/main.cpp
mjmagee991: Nice catch

cbolles (approved):
Looks good!

PR #60, opened 2022-04-10
Pulling us in a new direction

cbolles:
Add the ability to specify a pull direction (up, down, no pull) on GPIOs. Defaults to the previous hard coded value of `PULL_DOWN`.

### DIFF
--- a/include/EVT/io/GPIO.hpp
+++ b/include/EVT/io/GPIO.hpp
@@ -72,7 +72,7 @@ public:
      * @param[in] pin The pin for the GPIO instance to use.
      * @param[in] direction The directional flow of data.
      */
-    GPIO(Pin pin, Direction direction, Pull pull=Pull::PULL_DOWN);
+    GPIO(Pin pin, Direction direction, Pull pull = Pull::PULL_DOWN);
 
     /**
      * Set the direction of the pin.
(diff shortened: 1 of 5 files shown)


include/EVT/io/GPIO.hpp
```diff
@@ -63,7 +72,7 @@ class GPIO {
      * @param[in] pin The pin for the GPIO instance to use.
      * @param[in] direction The directional flow of data.
      */
-    GPIO(Pin pin, Direction direction);
+    GPIO(Pin pin, Direction direction, Pull pull = Pull::PULL_DOWN);
```
mjmagee991, line 75: In the summary of this PR, you say that the pull is supposed to default to NO_PULL, but here you have it as PULL_DOWN. Is this correct?

include/EVT/io/GPIO.hpp
```diff
@@ -63,7 +72,7 @@ class GPIO {
      * @param[in] pin The pin for the GPIO instance to use.
      * @param[in] direction The directional flow of data.
      */
```
mjmagee991, line 74: The documentation for several of these GPIO functions needs a new @param to reflect the new parameter.

mjmagee991 (changes requested):
Looks good. Just needs some documentation updates, and I had one clarifying question.

include/EVT/io/GPIO.hpp
cbolles: Good catch, should be `PULL_DOWN`

mjmagee991 (approved):
Looks good

zsteig (approved):

PR #63, opened 2022-07-10
I SPI Error Handling

AarC10:


mjmagee991:
Youre the best programmer I know

include/EVT/io/SPI.hpp
```diff
@@ -29,6 +29,17 @@ enum class Pin;
 
 class SPI {
 public:
+    /**
+     * Represents potential errors that may take place when using the SPI
```
mjmagee991, line 33: NIT: It doesn't just represent errors. It also represents the lack of errors. I'm not sure how this could be worded, but it could be slightly better.

samples/spi/SPI_With_Arduino/arduino_source/spi_slave_test/spi_slave_test.ino
```diff
@@ -1,7 +1,13 @@
-#define SCK_PIN   52
-#define MISO_PIN  50
-#define MOSI_PIN  51
-#define SS_PIN    53
+//#define SCK_PIN   52
```
mjmagee991, line 1: Delete unexplained pins

src/io/SPI.cpp
```diff
@@ -23,48 +23,103 @@ SPI::SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin) : CSPinsLen
     }
 }
 
-void SPI::write(uint8_t* bytes, uint8_t length) {
-    for (int i = 0; i < length; i++)
-        write(bytes[i]);
+SPI::SPIStatus SPI::write(uint8_t* bytes, uint8_t length) {
+    SPIStatus status;
```
mjmagee991, line 27:
```suggestion
    SPIStatus status = SPIStatus::OK;
```

src/io/SPI.cpp
```diff
@@ -23,48 +23,103 @@ SPI::SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin) : CSPinsLen
...
+
+    for (int i = 0; i < length; i++) {
+        status = write(bytes[i]);
+        if (status != SPIStatus::OK)
+            return status;
+    }
+
+    return SPIStatus::OK;
```
mjmagee991, line 35:
```suggestion
    return status;
```

src/io/SPI.cpp
```diff
@@ -23,48 +23,103 @@ SPI::SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin) : CSPinsLen
...
+SPI::SPIStatus SPI::writeReg(uint8_t device, uint8_t reg, uint8_t byte) {
+    bool transmitSuccess = false;
+    SPIStatus status;
+
     if (startTransmission(device)) {
-        write(reg);
-        write(byte);
-        endTransmission(device);
+        status = write(reg);
+        if (status != SPIStatus::OK)
```
mjmagee991, line 55: Make this into a macro

mjmagee991 (changes requested):
Couple stylistic issues, but overall pretty good.

cmake/evt-core_compiler.cmake
```diff
@@ -8,6 +8,7 @@ if(NOT DEFINED ENV{GCC_ARM_TOOLS_PATH})
     message(WARNING
             "Set your environment variables you frickin' hecker."
             "   --Shane Snover"
+
```
mjmagee991, line 11: Why did you add this?

docs/source/index.rst
```diff
@@ -46,7 +46,7 @@ project is built using CMake and requires an environment variable containing
 the location of the GCC ARM toolchain. The README will go over setting the
 environment variable and building EVT-core.
 
-Several samples are included in EVT-core directly. As a libary, EVT-core can
+Several samples are included in EVT-core directly. As a libzzary, EVT-core can
```
mjmagee991, line 49: This is not much better

src/io/SPI.cpp
```diff
@@ -1,5 +1,9 @@
 #include <EVT/io/SPI.hpp>
 
+#define SPI_CHECK                \
```
mjmagee991, line 3: I learned about C++ error handling on my co-op, and I think this should look something like the below:
```suggestion
#define SPI_CHECK(functionCall) SPIStatus status = functionCall; if(status != SPIStatus::OK) return status
```
If you want it to be able to be used more than once in  a function, you need to do something like this: https://stackoverflow.com/questions/1082192/how-to-generate-random-variable-names-in-c-using-macros

src/io/SPI.cpp
```diff
@@ -23,48 +27,93 @@ SPI::SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin) : CSPinsLen
...
 }
 
-void SPI::write(uint8_t* bytes, uint8_t length) {
-    for (int i = 0; i < length; i++)
-        write(bytes[i]);
+SPI::SPIStatus SPI::write(uint8_t* bytes, uint8_t length) {
+    SPIStatus status = SPIStatus::OK;
+
+    for (int i = 0; i < length; i++) {
+        status = write(bytes[i]);
```
mjmagee991, line 34: With the previous change, this line would change like this:
```suggestion
        SPI_CHECK(write(bytes[i]));
```

src/io/SPI.cpp
```diff
@@ -23,48 +27,93 @@ SPI::SPI(GPIO* CSPins[], uint8_t pinLength, Pin sckPin, Pin mosiPin) : CSPinsLen
...
+        SPI_CHECK
+    }
+
+    return status;
```
mjmagee991, line 38: And this would become
```suggestion
    return SPIStatus::OK;
```

mjmagee991 (changes requested):
Sorry if these changes sort of contradict the previous changes, but I think this way of handling the errors makes a bit more sense.

src/io/SPI.cpp
```diff
@@ -1,5 +1,10 @@
 #include <EVT/io/SPI.hpp>
 
+#define SPI_CHECK(functionCall)  \
```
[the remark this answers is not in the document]
mjmagee991: Basically a new 'status' variable would be created for each call to the macro, which eliminates the need for people to instantiate `SPIStatus status = SPIStatus::OK` before using the macro. This makes it completely self-contained, which I believe makes it much more readable than the way it is now. We should talk about this in person next chance we get.

src/io/SPI.cpp
```diff
@@ -1,5 +1,13 @@
 #include <EVT/io/SPI.hpp>
 
+#define CONCAT(a, b) CONCAT_INNER(a, b)
+#define CONCAT_INNER(a, b) a##b
+
+#define SPI_CHECK(functionCall, statusID)              \
```
mjmagee991, line 6: I did some testing and came up with this solution:
```suggestion
#define SPI_CHECK(functionCall)                        \
    SPIStatus CONCAT(status, __LINE__) = functionCall; \
    if (CONCAT(status, __LINE__) != SPIStatus::OK)     \
    return CONCAT(status, __LINE__)
```

This would allow you to eliminate the `statusID`, making the macro completely self-contained. From what I've seen, I think this is the ideal solution.

mjmagee991:
[SPI Error Handling](https://trello.com/c/lAmG6PRc/29-spi-error-handling)

mjmagee991 (approved):
Looks great

ActuallyZach (approved):

PR #64, opened 2022-10-01
The Mr.Raw

tmb7852:
This change added the ability for thermistors to retrieve the raw voltage data from the ADC.

### DIFF
--- a/samples/can/back_and_forth/main.cpp
+++ b/samples/can/back_and_forth/main.cpp
@@ -50,8 +50,6 @@ int main() {
     if (result != IO::CAN::CANStatus::OK) {
         uart.printf("Failed to connect to the CAN network\r\n");
         return 1;
-    }else{
-        uart.printf("Connection Successful!\r\n");
     }
 
     while (true) {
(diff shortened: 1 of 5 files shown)


samples/can/back_and_forth/main.cpp
```diff
@@ -50,6 +50,8 @@ int main() {
     if (result != IO::CAN::CANStatus::OK) {
         uart.printf("Failed to connect to the CAN network\r\n");
         return 1;
+    } else {
+        uart.printf("Connection Successful!\r\n");
```
AarC10, line 54: Else statement is redundant here. If the CANStatus is not ok, the printf statement would never reach anyways

AarC10 (changes requested):
Good job! You can just remove some redundant code and you should be good to go.

samples/can/back_and_forth/main.cpp
ActuallyZach: Is there any detriment to having it? I see how it is redundant but it could be helpful if you are looking for debug information.

AarC10: Not really a detriment besides just being a style issue. The compiler would optimize the same anyways. Not understanding fully what you mean by looking for debug information.

tmb7852: Issue should be resolved. I was using that as tester code to check if everything was up and running. Should be good now, if not let me know.

AarC10 (approved):

mjmagee991 (approved):
Ok

PR #68, opened 2022-11-20
Don't blow your Stack

mjh9585:
- Added macro tests to allow the queue sizes to be adjusted with a simple compile flag. 
- Also cut the CANOpen queue size in half to significantly reduce stack usage.

_*Note: Each CANMessage in the queue is 16 bytes so the original 300 used 4,800 bytes of stack (a literal third). This fix reduces this to 2,400 bytes. I would probably recommend dropping this further to 50 elements or 800 bytes_

### DIFF
--- a/include/EVT/io/CANopen.hpp
+++ b/include/EVT/io/CANopen.hpp
@@ -14,7 +14,10 @@
 #include <EVT/io/types/CANMessage.hpp>
 #include <EVT/utils/types/FixedQueue.hpp>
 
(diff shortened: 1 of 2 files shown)


mjmagee991 (approved):
Looks great!

mjmagee991:
[CANopen Stack Overflow](https://trello.com/c/uw06aAEx/9-canopen-stack-overflow)

AarC10 (approved):

PR #69, opened 2022-11-29
I c display firmware

ActuallyZach:
Updated the LCD firmware currently present in EVT-Core. The lcd is now capable of displaying in a 9 segment display mode, a single bit map mode and a multi bitmap mode. 6 samples were also added to show different ways of using the LCD.

Additions
- Text Display
- Individual Bitmap display
- Improved drawing ability of the display
- Converted pages and columns to near X and Y values.
- Documented the class and functions


include/EVT/dev/BitmapFonts.h
```diff
@@ -0,0 +1,250 @@
...
+        { 0x1B,0x1E,0x17,0x0 },// Character: zcaron
+        { 0x0,0x0,0x0,0x0 },// Character: uni0EA4
+        { 0x0,0x0,0x0,0x0 },// Character: uni13A0
+        { 0x4,0x0,0x0,0x0 },// Character: bullet
+        { 0x4,0x0,0x4,0x0 },// Character: ellipsis
+        { 0xE,0x1F,0x17,0x0 },// Character: Euro
+    };
+};
+
+#endif//EVT_BITMAPFONTS_H
```
AarC10, line 250: Newline at the end here

include/EVT/dev/LCD.hpp
```diff
@@ -74,22 +77,86 @@ class LCD {
...
      */
     void initLCD();
 
 private:
-    /** register select pin for the LCD */
+    /** The total width of the screen */
+    static const int screenSizeX = 128;
+
+    /** The total height of the screen */
+    static const int screenSizeY = 64;
```
AarC10, line 127: Is there a reason these cant be unsigned integers too?

samples/lcd/basic/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+        lcd.clearArea(16, 8, 7, 0);
+        const char* dst = std::to_string(number).c_str();
+        lcd.writeText(dst, 7, 0, true);
+
+        number++;
+        time::wait(500);
+    }
+
+    return 0;
+}
```
AarC10, line 81: Newline at end

samples/lcd/lcd-alphabet/main.cpp
```diff
@@ -0,0 +1,53 @@
+//
+// Created by Zachary Lineman on 11/14/22.
+//
```
AarC10, line 3: Assuming this is CLion autogenerated. Don't think this follows our style guidelines

samples/lcd/lcd-ball/main.cpp
```diff
@@ -0,0 +1,74 @@
+//
+// Created by Zachary Lineman on 11/14/22.
+//
```
AarC10, line 3: Same thing mentioned previously

samples/lcd/lcd-counter/main.cpp
```diff
@@ -0,0 +1,62 @@
+//
+// Created by Zachary Lineman on 11/14/22.
+//
```
AarC10, line 3: :)

src/dev/LCD.cpp
```diff
@@ -23,71 +39,221 @@ void LCD::commandWrite(uint8_t data) {
...
-    this->commandWrite(0x40);          //Display start address + 0x40
-    for (int i = 0; i < 8; i++) {      //64 pixel display / 8 pixels per page = 8 pages
-        this->commandWrite(page);      //send page address
-        this->commandWrite(0x10);      //column address upper 4 bits + 0x10
-        this->commandWrite(0x00);      //column address lower 4 bits + 0x00
-        for (int j = 0; j < 128; j++) {//128 columns wide
-            this->dataWrite(0x00);     //write clear pixels
-            bitMap++;
+void LCD::clearLCD() {
+    uint8_t i, j;
```
AarC10, line 67: I would suggest assigning a value here, or just assign them inside your loop since it does not look like you use these value outside of the loop

src/dev/LCD.cpp
```diff
@@ -23,71 +39,221 @@ void LCD::commandWrite(uint8_t data) {
...
 void LCD::displayMap(const uint8_t* bitMap) {
+    uint8_t i, j;
     uint8_t page = 0xB0;
-    this->commandWrite(0xAE);          //Display OFF
-    this->commandWrite(0x40);          //Display start address + 0x40
-    for (int i = 0; i < 8; i++) {      //64 pixel display / 8 pixels per page = 8 pages
-        this->commandWrite(page);      //send page address
-        this->commandWrite(0x10);      //column address upper 4 bits + 0x10
-        this->commandWrite(0x00);      //column address lower 4 bits + 0x00
-        for (int j = 0; j < 128; j++) {//128 columns wide
```
AarC10, line 67: Same thing mentioned previously. Sort of curious why you do it this way?

src/dev/LCD.cpp
```diff
@@ -23,71 +39,221 @@ void LCD::commandWrite(uint8_t data) {
...
+
+        page++;// After 128 columns, go to next page
+    }
+    this->commandWrite(0xAF);// Finish writing
+}
+
+void LCD::writeText(const char* text, uint8_t page, uint8_t column, bool wrapText) {
+    for (uint8_t x = 0; x < strlen(text); x++) {
+        // Get the ASCII value of the character.
+        uint8_t fontIndex = (int) ((unsigned char) text[x]);
```
AarC10, line 172: C++ style cast might be better here so you can check the cast at compile time

AarC10 (changes requested):
Nice work. Just a few nitpicks I found. I think I got most of them and making other comments would be redundant. Just remember to follow the team's style guidelines mostly.

include/EVT/dev/LCD.hpp
ActuallyZach: There is not, just swapped them

src/dev/LCD.cpp
ActuallyZach: If I remember correctly this was just the way it was when I got to the code and I never changed it. I moved them all into the loops.

AarC10 (changes requested):

samples/lcd/lcd-segment/main.cpp
```diff
@@ -10,7 +6,7 @@
 #include <string>
```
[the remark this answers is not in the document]
ActuallyZach: The reason that is in the sample code is because it is using to_string to convert a number into a string. Is there another way to do this without using <string>?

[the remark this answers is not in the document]
AarC10: You can statically allocate a buffer and then use sprintf or snprintf.
`char buffer[SOME LARGE ENOUGH SIZE] = {}`
`sprintf(buffer, "%d", (SOME NUMBER));`
`snprintf(buffer, (MAX CHARS), "%d", (SOME NUMBER));`

AarC10 (approved):

include/EVT/dev/LCD.hpp
```diff
@@ -74,22 +77,86 @@ class LCD {
...
+     * @param column the column to write the text too. Range 0-127.
+     * @param wrapText whether or not the text should be wrapped around the edge of the screen.
+     */
+    void writeText(const char* text, uint8_t page, uint8_t column, bool wrapText);
+
+    /**
+     * Set the default section titles to be displayed.
+     * @param newSectionTitles an array of section titles to display.
+     */
+    void setDefaultSections(char* newSectionTitles[9]);
```
mjmagee991, line 103: Having these functions here is okay for now, but at some point in the future, we should find a way to abstract them into a different class. I'm not sure what exactly that would look like, but we can talk about it next semester.

include/EVT/dev/LCD.hpp
```diff
@@ -74,22 +77,86 @@ class LCD {
...
+
     /**
      * Initializes the LCD for operation (must be called to use the LCD)
      */
     void initLCD();
 
 private:
-    /** register select pin for the LCD */
+    /** The total width of the screen */
+    static const uint8_t screenSizeX = 128;
```
mjmagee991, line 124: Likewise, we should try to abstract some of these values away, so the LCD class can be generic for use with other LCDs. It's not a priority right now, but it's something to think about.

src/dev/LCD.cpp
```diff
@@ -1,15 +1,31 @@
+#include "EVT/dev/BitmapFonts.h"
```
mjmagee991, line 1: This should be a `<>` include

samples/CMakeLists.txt
```diff
@@ -20,4 +20,9 @@ add_subdirectory(rtc_timer)
 add_subdirectory(thermistor)
 add_subdirectory(timer)
 add_subdirectory(spi)
-add_subdirectory(lcd)
+add_subdirectory(lcd/basic)
```
mjmagee991, line 23: All of these samples were important for experimenting with the LCD, but now that we have the LCD working the way we want, do you think we need all of them? Are there any you could consolidate?

src/dev/LCD.cpp
```diff
@@ -23,71 +39,217 @@ void LCD::commandWrite(uint8_t data) {
...
-    this->dataWrite(data);//writes 8 vertical bits based on value between 0-255 based on bits set ex: 01001100(0x4C) is |WHITE|
-                          //                                                                                                 |BLACK|
-                          //                                                                                                 |WHITE|
-                          //                                                                                                 |WHITE|
-                          //                                                                                                 |BLACK|
-                          //                                                                                                 |BLACK|
-                          //                                                                                                 |WHITE|
-                          //                                                                                                 |WHITE|
+void LCD::driveColumn(uint8_t page, uint8_t colUp, uint8_t colLow, uint8_t data) {
+    this->commandWrite(0xAE);// Display OFF
```
mjmagee991, line 43: Is this necessary? Could this be causing the flashing we're seeing?

src/dev/LCD.cpp
```diff
@@ -23,71 +39,217 @@ void LCD::commandWrite(uint8_t data) {
...
+        page++;// After 128 columns, go to next page
     }
     this->commandWrite(0xAF);
 }
 
+void LCD::clearArea(uint8_t width, uint8_t height, uint8_t page, uint8_t column) {
+    // Calculate the correct column address from an 8-bit number.
+    // The issue is that the display takes 4 bit numbers, instead of 8.
+    // So you need to split it into two 4 bit numbers.
+    uint8_t columnUpperAddress = column;
```
mjmagee991, line 87: This can be done a bit more simply by using bitwise AND. For example, you can get the upper address with `uint8_t columnUpperAddress = (column & 0xf0) >> 4;`

include/EVT/dev/LCD.hpp
```diff
@@ -74,22 +77,86 @@ class LCD {
      */
     void displayMap(const uint8_t* bitMap);
```
mjmagee991, line 78: `displayMap` and `displayBitmap` have very similar names. Can you change them to make their usage a bit clearer? If you think their behavior is similar enough, you could also overload the function, so they have the same name.

src/dev/LCD.cpp
```diff
@@ -23,71 +39,217 @@ void LCD::commandWrite(uint8_t data) {
...
     }
     this->commandWrite(0xAF);
 }
 
-void LCD::initLCD() {
+void LCD::displayBitMap(uint8_t* bitMap, uint8_t bitMapWidth, uint8_t bitMapHeight, uint8_t page, uint8_t column) {
+    // Calculate the correct column address from an 8-bit number.
+    // The issue is that the display takes 4 bit numbers, instead of 8.
+    // So you need to split it into two 4 bit numbers.
+    uint8_t columnUpperAddress = column;
```
mjmagee991, line 136: This calculation can be done more simply in the same way as in `clearArea`.

src/dev/LCD.cpp
```diff
@@ -23,71 +39,217 @@ void LCD::commandWrite(uint8_t data) {
...
+
+        page++;// After 128 columns, go to next page
+    }
+    this->commandWrite(0xAF);// Finish writing
+}
+
+void LCD::writeText(const char* text, uint8_t page, uint8_t column, bool wrapText) {
+    for (uint8_t x = 0; x < strlen(text); x++) {
+        // Get the ASCII value of the character.
+        uint8_t fontIndex = int(uint8_t(text[x]));
```
mjmagee991, line 168: Why are you casting to `uint8_t` to `int`, and back to `uint8_t`?

src/dev/LCD.cpp
```diff
@@ -23,71 +39,217 @@ void LCD::commandWrite(uint8_t data) {
...
+
+        // Create the character that we need to write to the screen.
+        unsigned char characterMap[4] = {
+            BitmapFont::font4x6[fontIndex][0],
```
mjmagee991, line 172: This is fine for now, but we should abstract this behavior to work with other size fonts in the future.

mjmagee991 (changes requested):
Overall, this is a great PR. You've been doing great work with the LCD driver, and I'm glad to see it's coming together. I do think we should have tried to break this down a bit more into multiple PRs because there is so much code here, but we can work on that in the future. I found a few small things that need to be changed and had a few questions about why you implemented things the way you did

include/EVT/dev/LCD.hpp
```diff
@@ -6,7 +6,7 @@
 #include <EVT/io/UART.hpp>
 #include <EVT/io/pin.hpp>
 #include <EVT/utils/time.hpp>
-#include <stdint.h>
+#include <cstdint>
```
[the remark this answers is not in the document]
ActuallyZach: I changed it because Clion warned that stdint.h is a depreciated header and that cstdint is its replacement. I can swap it bakc if you would like.

include/EVT/dev/LCD.hpp
```diff
@@ -38,34 +37,38 @@ class LCD {
...
-    * @param[in] data command to write to the LCD
-    */
+     * Writes commands to the LCD.
+     * These commands are used to tell the display how to handle the data written by dataWrite();
+     * @param data the command to write to the LCD
+     */
     void commandWrite(uint8_t data);
 
     /**
-     * Drives a particular pixel on the LCD
```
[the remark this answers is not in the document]
ActuallyZach: Oops, that drives a whole column. Fixed that documentation

include/EVT/dev/LCD.hpp
ActuallyZach: Okay sounds good. I definitely agree that it should be abstracted outside of here in some form

samples/CMakeLists.txt
ActuallyZach: We could totally consolidate a few of them. Specifically Alphabet and ball which are combined into one in the basic example. We could also combine counter and segment as we can just have one of the segments count in that sample.

Should I go through and consolidate these samples?

src/dev/LCD.cpp
ActuallyZach: That is a possibility, I never tested without it so there is a good shot that is the flashing we are seeing. This is something we should check out when next possible.

samples/lcd/lcd-counter/main.cpp
```diff
@@ -0,0 +1,60 @@
...
+constexpr uint8_t deviceCount = 1;
+
+IO::GPIO* devices[deviceCount];
+
+int main() {
+    // Initialize system
+    IO::init();
+
+    // Setup UART
+    IO::UART& uart = IO::getUART<IO::Pin::PB_10, IO::Pin::PB_11>(9600);
```
mjmagee991, line 27: Can you fix the UART pins for these targets as well?

include/EVT/dev/BitmapFonts.hpp
```diff
@@ -0,0 +1,252 @@
+/**
+* BitMapFont is a class that holds a representation of a bitmap font. The class can hold multiple fonts, however
+ * currently it is only holding a single 4x6 font.
```
mjmagee991, line 3: This comment should be  immediately before the class declaration, and the header guards should be at the top. This one has to be near the class instead of at the top of the file, so the documentation generator can find it.

mjmagee991 (changes requested):
Great improvements. I just found two more nitpicks, but it looks great otherwise

src/dev/LCD.cpp
ActuallyZach: Agreed I think we need to come up with a system for picking either setting the font per text writing or for the whole device

mjmagee991 (approved):
Two little issues, but looks good to go

PR #71, opened 2023-01-08
Would You Watch my Dog?

mjmagee991:
Renamed IWDG, RTC, and Timer implementation classes to indicate that they work on all STM32f3xx series MCUs, not just the 302
Cleaned up some documentation and formatting for those classes


mjmagee991:
[Platform Device Support on STM32f334](https://trello.com/c/IsCF2rnA/56-platform-device-support-on-stm32f334)

src/dev/platform/f3xx/IWDGf3xx.cpp
```diff
@@ -0,0 +1,23 @@
...
+// the ratio of counter ticks to milliseconds should be about 1:8. In testing,
+// we found this to have an error of about 1 second
+IWDGf3xx::IWDGf3xx(uint32_t ms) : halIWDG{
+    IWDG1,
+    {
+        ms / 8 - 1,
+        ms / 8 - 1,
+        IWDG_PRESCALER_256,
+    },
+} {
```
mjh9585, line 15: I have never seen this format before. Just want to confirm that this works correctly.

mjh9585 (approved):
Other than the comments and questions I made, looks good.

samples/log/main.cpp
```diff
@@ -28,7 +28,7 @@ int main() {
     // If timestamps aren't needed, don't set the logger's clock
     log::LOGGER.setUART(&uart);
     log::LOGGER.setLogLevel(log::Logger::LogLevel::INFO);
-    auto* rtc = new DEV::RTCf302x8();
+    auto* rtc = new DEV::RTCf3xx();
```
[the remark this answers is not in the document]
AarC10: I think the auto is fine here imo. The compiler should still be able to make the correct determination.

AarC10, line 31: My main concern here is why are we dynamically allocating the rtc?

src/dev/platform/f3xx/RTCf3xx.cpp
```diff
@@ -1,24 +1,27 @@
...
+    {
+        // Numbers generated from STMCubeMX
+        RTC_HOURFORMAT_24,
+        127,
+        255,
+        RTC_OUTPUT_DISABLE,
+        RTC_OUTPUT_TYPE_PUSHPULL,
+        RTC_OUTPUT_POLARITY_HIGH,
+    },
+} {
```
AarC10, line 20: Like what Matt mentioned above, this formatting looks a little scuffed

AarC10 (changes requested):

src/dev/platform/f3xx/IWDGf3xx.cpp
mjmagee991: Yeah, this works. It's called an initializer list.

samples/log/main.cpp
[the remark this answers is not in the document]
mjmagee991: I think it makes sense to have a strongly typed RTC here. That's the way we intend for it to be used, and I want it to be explicit, so people who look at this for reference in the future see it used correctly.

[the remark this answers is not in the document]
mjmagee991: Realistically, what we actually need is a factory method that returns the correct concrete implementation of RTC, but I'm not sure how to do that without dynamic allocation.

src/dev/platform/f3xx/RTCf3xx.cpp
mjmagee991: I'm not really sure how this works, but CLion really wants me to do this, and it seems to make stuff a little more explicit, so I'm okay with it

AarC10 (approved):

zsteig (commented):
Just some questions about TODOs (Not sure why this didn't upload sooner)

ActuallyZach (approved):

PR #74, opened 2023-01-31
Flash me one last time

ActuallyZach:
Removed the weird flashing from the HUDL. Did this by simply removing the DISPLAYOFF signal that was sent every time something was drawn to the screen.

Tested all samples and the HUDL still functions like expected.


mjmagee991 (approved):
Looks good. Nice job

mjmagee991 (approved):
Looks good

AarC10 (approved):

PR #75, opened 2023-02-14
I too (2) can now see (C) multiple bytes

pro-sper:
readReg() method modified for multibyte i2c reads

### DIFF
--- a/src/io/I2C.cpp
+++ b/src/io/I2C.cpp
@@ -42,17 +42,17 @@ I2C::I2CStatus I2C::writeReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
 /**
  * Reading from a register with a non-8 byte address and a non-8 byte
  * response usually involves writing out the register address in groups
- * of 8 bytes starting with the MSB. Reading the response is also done
+ * of 8 bytes starting with the MSB. Reading the response is also done,
  * starting with the LSB.
  */
 I2C::I2CStatus I2C::readReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
                             uint8_t* bytes, uint8_t length) {
     // Write out register address
-    for (int i = 0; i < regLength; i++)
-        I2C_RETURN_IF_ERR(write(addr, reg[i]));
+    I2C_RETURN_IF_ERR(write(addr, reg, length));
+
     // Read in response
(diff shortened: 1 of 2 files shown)


pro-sper:
[I2C Read Multiple Bytes](https://trello.com/c/wik1gcJI/120-i2c-read-multiple-bytes)

src/io/I2C.cpp
```diff
@@ -42,17 +42,17 @@ I2C::I2CStatus I2C::writeReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
...
- * of 8 bytes starting with the MSB. Reading the response is also done
+ * of 8 bytes starting with the MSB. Reading the response is also done,
  * starting with the LSB.
  */
 I2C::I2CStatus I2C::readReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
                             uint8_t* bytes, uint8_t length) {
     // Write out register address
-    for (int i = 0; i < regLength; i++)
-        I2C_RETURN_IF_ERR(write(addr, reg[i]));
+    I2C_RETURN_IF_ERR(write(addr, reg[0]));
```
mjmagee991, line 51: This write call is only writing 1 byte of the register address. You should use the write override that writes multiple bytes

src/io/I2C.cpp
```diff
@@ -42,17 +42,17 @@ I2C::I2CStatus I2C::writeReg(uint8_t addr, uint8_t* reg, uint8_t regLength,
...
+
     // Read in response
-    for (int i = 0; i < length; i++)
-        I2C_RETURN_IF_ERR(read(addr, &bytes[i]));
+    I2C_RETURN_IF_ERR(read(addr, &bytes[0], length));
```
mjmagee991, line 54:
```suggestion
    I2C_RETURN_IF_ERR(read(addr, bytes, length));
```
Deleting these characters doesn't actually change the behavior of this call. With the ampersand and brackets, you're passing the address of the first element in the array, which is what `bytes` already is

mjmagee991 (changes requested):
A couple minor changes

mjmagee991 (approved):
Nice

ishan-renjen (approved):

ActuallyZach (approved):

PR #77, opened 2023-03-03
U ARen'T Reading the Documentation

mjmagee991:
Enabled UART pin swapping

### DIFF
--- a/include/EVT/io/platform/f3xx/UARTf3xx.hpp
+++ b/include/EVT/io/platform/f3xx/UARTf3xx.hpp
@@ -19,7 +19,7 @@ public:
      * @param[in] rxPin THe UART RX pin.
      * @param[in] baudrate The baudrate to operate the UART with/
      */
-    UARTf3xx(Pin txPin, Pin rxPin, uint32_t baudrate);
+    UARTf3xx(Pin txPin, Pin rxPin, uint32_t baudrate, bool isSwapped);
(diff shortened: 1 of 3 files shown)


jlb1854:
I'm glad we have the hardware in place to fix this issue. Thank you for implementing this software feature.

jlb1854 (approved):
Great work!

jlb1854 (changes requested):
Is this the best implementation you can have? This may cause confusion as it affects the entire repository, can you pass this as an argument to the constructor in the same manner as the pins? You could also implement this as part of the switch case that determines which halUart object you are referencing with the pins passed to the constructor.

include/EVT/io/platform/f3xx/UARTf3xx.hpp
```diff
@@ -19,7 +19,7 @@ class UARTf3xx : public UART {
      * @param[in] rxPin THe UART RX pin.
      * @param[in] baudrate The baudrate to operate the UART with/
      */
```
ActuallyZach, line 21: What does isSwapped do? There should be documentation for it.

ActuallyZach (changes requested):

ActuallyZach (approved):
Looks good!

jlb1854:
Great job!

jlb1854 (approved):

PR #82, opened 2023-05-06
Bilingual LCDs

ActuallyTaylor:
Added a new font to the LCDs with the ability to write small and large text

Changes
- Added new font
- Updated samples
- Removed static section height parameter


include/EVT/dev/LCD.hpp
```diff
@@ -88,11 +86,11 @@ class LCD {
      * Clears only a certain area on the display screen.
      *
      * @param[in] width the width in pixels of the area to clear. Range: 0-127
-     * @param[in] height the height in pixels of the area to clear. Range: 0-63
+     * @param[in] numbPages the height in pixels of the area to clear. Range: 0-63
```
mjmagee991, line 89: This documentation should be updated to reflect the purpose of the new variable. Also, for my sanity, can we call this `numPages`, instead of `numbPages`?

include/EVT/dev/LCD.hpp
```diff
@@ -106,7 +104,7 @@ class LCD {
      *
      * @param bitMap[in] the bitmap to display.
      * @param bitMapWidth[in] the width of the bitmap in pixels.
-     * @param bitMapHeight[in] the height of the bitmap in pixels.
+     * @param numbPages[in] the number of pages that the bitmap requires to draw.
```
mjmagee991, line 107: This documentation doesn't agree with the function header

include/EVT/dev/LCD.hpp
```diff
@@ -120,7 +118,17 @@ class LCD {
      * @param column[in] the column to write the text too. Range 0-127.
      * @param wrapText[in] whether or not the text should be wrapped around the edge of the screen.
      */
-    void writeText(const char* text, uint8_t page, uint8_t column, bool wrapText);
+    void writeSmallText(const char* text, uint8_t page, uint8_t column, bool wrapText);
```
mjmagee991, line 121: I can see that the implementation is relatively different for writing small and large text, but I feel like, from an interface standpoint, it would make more sense to have a single `writeText()` method and then set the font size somewhere else. I think I would prefer to have weird behavior if used improperly than make the application code explicitly say the font size every time it writes text on the screen. Does that make sense?

mjmagee991 (changes requested):
Some thoughts on the interface

include/EVT/dev/LCD.hpp
ActuallyTaylor: That makes sense, I just condensed the function into a single `writeText()` that allows us to select the font size during the call

mjmagee991 (approved):
Nice work

src/dev/LCD.cpp
```diff
@@ -255,6 +212,65 @@ void LCD::setTextForSection(uint8_t section, const char* text) {
...
+        case SMALL: {
+            unsigned char characterMap[4] = {
+                BitmapFont::font4x6[characterIndex][0],
+                BitmapFont::font4x6[characterIndex][1],
+                BitmapFont::font4x6[characterIndex][2],
+                BitmapFont::font4x6[characterIndex][3],
+            };
+
+            // Display the character bit map at the calculated page and column.
+            displayBitMapInArea(characterMap, characterWidth, characterPages, page, column);
```
mjh9585, line 263: Is the use of characterMap necessary? Could you not just call `displayBitMapInArea(BitmapFont::font4x6[characterIndex], characterWidth, characterPages, page, column);`

mjh9585 (approved):
Overall looks good! Just one question I had for you to resolve.

src/dev/LCD.cpp
ActuallyTaylor: Thats a good catch, I don't think it is necessary for that part of the function. I think for the `writeLargeText` function it is still going to be necessary because 3 bytes of 0s need to be appended. But if there is some other way to do it I think it would be a good idea

PR #83, opened 2023-09-29
🐛 :hammer:

aclowmclaughlin:
Fixed a bug in I2C readReg()

It was using the length of the data output as the number of registers to read from, which was creating situations where it was accessing random memory locations


ActuallyTaylor (approved):

mjmagee991 (approved):
Nice catch

PR #84, opened 2023-10-05
Nice dog, innit?

mjmagee991:
Added init() function to start the running of the watchdog timer sometime after the class has been instantiated
Needed for the BMS to work properly
(Also fixed a mistake I made when I modified the class a year ago, which completely broke it)

### DIFF
--- a/include/EVT/dev/IWDG.hpp
+++ b/include/EVT/dev/IWDG.hpp
@@ -10,6 +10,11 @@ namespace EVT::core::DEV {
 class IWDG {
 public:
     /**
+     * Initializes and activates the watchdog timer.
+     */
+    virtual void init() = 0;
+
+    /**
      * Refreshes the watchdog timer.
      */
     virtual void refresh() = 0;
--- a/src/dev/platform/f3xx/IWDGf3xx.cpp
+++ b/src/dev/platform/f3xx/IWDGf3xx.cpp
@@ -12,12 +12,17 @@ IWDGf3xx::IWDGf3xx(uint32_t ms) : halIWDG{
         ms / 8 - 1,
         ms / 8 - 1,
     },
-} {
+} { }
+
+void IWDGf3xx::init() {
     HAL_IWDG_Init(&halIWDG);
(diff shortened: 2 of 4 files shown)


src/dev/platform/f3xx/IWDGf3xx.cpp
```diff
@@ -12,12 +12,17 @@ IWDGf3xx::IWDGf3xx(uint32_t ms) : halIWDG{
         ms / 8 - 1,
         ms / 8 - 1,
     },
-} {
+} {}
+
+void IWDGf3xx::init() {
     HAL_IWDG_Init(&halIWDG);
+    isActive = true;
```
AarC10, line 19: Should any error handling be done here? I.E check HAL return code and return our own EVT-Core status code

AarC10 (approved):
Looks good, but suggesting that error handling be done for initializing the watchdog instead of just expecting that it will always be successful.

include/EVT/dev/IWDG.hpp
```diff
@@ -9,6 +9,11 @@ namespace EVT::core::DEV {
  */
 class IWDG {
 public:
+    /**
+     * Initializes and activates the watchdog timer.
+     */
+    virtual void init() = 0;
```
ActuallyTaylor, line 15: Why is this initialized to 0? Is it because it is virtual and get's implemented by individual boards?

ActuallyTaylor (approved):

include/EVT/dev/IWDG.hpp
mjmagee991: Yes, that's the way C++ likes virtual methods to be initialized, as I understand it

src/dev/platform/f3xx/IWDGf3xx.cpp
mjmagee991: There probably should be some error handling, but that is going to have to be addressed across EVT-core, so I'd rather not hold this PR up for it

PR #85, opened 2023-10-05
CAN for 100 secs

mjmagee991:
Added variable CAN Timeout, so boards can have different timeouts if needed


ActuallyTaylor (approved):
This looks like a good implementation of a timeout for CAN!

PR #87, opened 2023-10-21
Don't let the encoder drive the bus

aclowmclaughlin:
Created a driver for a 2-bit encoder, currently used on the charge controller. Because it has a low range, the code internally keeps track of the current direction it is rotating in and assumes you will still be rotating in that direction, until several updates have occurred without a change.
The current version of the code does not use interrupts and so may be slower to update if the loop it is in is doing some other processing.


include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+namespace EVT::core::DEV {
+class Encoder {
+public:
+    /**
+     * Constructs an Encoder instance. Call update() within the main loop to update encoder values.
+     * @param a a pin of the encoder
+     * @param b b pin of the encoder
+     * @param range range of the encoder positions (should be positive)
+     * @param initialPosition initial position the encoder is in
```
AarC10, line 14: Specify whether this is [in] or [out]

aclowmclaughlin:
[Encoder Driver](https://trello.com/c/VtgG1Uyv/129-encoder-driver)

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+    Encoder(IO::GPIO& a, IO::GPIO& b, int64_t range, int64_t initialPosition);
+
+    /**
+     * An Enum that represents the direction the encoder is being spun in.
+     * The direction should reset to Static after a couple of reads with no changes.
+     */
+    enum Direction {
+        Left = -1,
+        Static = 0,
+        Right = 1
```
AarC10, line 25: Should be fully capitalized

AarC10 (changes requested):

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+     * returns the noChangeCounter value;
+     * @return the noChangeCounter value
+     */
+    uint8_t getNoChangeCounter();
+
+    int8_t getRelativePosition();
+
+private:
+    ///GPIO pin a
+    IO::GPIO& a;
```
ActuallyTaylor, line 50: Are there more descriptive names to use other than just a and b? Or is this the standard put forth in a datasheet?

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+    ///GPIO pin b
+    IO::GPIO& b;
+
+    ///position can be between = [-range, range]
+    int64_t range;
+    ///the current absolute position of the encoder
+    int64_t position;
+    ///the current direction of travel of the encoder
+    Direction currentDirection;
+    ;///counter for how many updates have occured since the encoder was moved
```
ActuallyTaylor, line 60: I'm assuming this ; is a mistake and shouldn't be here

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+     * sets position to newPosition, capping it if the new value exceeds the range.
+     * @param newPosition the new value for position
+     * @return if position was capped
+     */
+    bool setPosition(int64_t newPosition);
+
+};// namespace EVT::core::DEV
+}// namespace EVT::core::DEV
+
+#endif
```
ActuallyTaylor, line 112:
```suggestion
#endif

```

samples/encoder/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+        //PRINT VALUES (only enable one at a time)
+        //uart.printf("\r Encoder Change: %d       ", change);
+        //uart.printf("\rRelative Pos: %d          ", currentRelPos);
+        uart.printf("\rPosition: %d         ", position);
+        //uart.printf("\rNoChangeCounter: %d          ", noChangeCounter);
+
+        //The wait simulates a loop that is doing other processing, because that will affect how often the output is read
+        time::wait(300);
+    }
+}
```
ActuallyTaylor, line 66:
```suggestion
}

```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+uint8_t Encoder::getNoChangeCounter() {
+    return noChangeCounter;
+}
+
+int8_t Encoder::getRelativePosition() {
+    return currentRelPos;
+}
+
+}// namespace EVT::core::DEV
```
ActuallyTaylor, line 152:
```suggestion
}// namespace EVT::core::DEV

```

ActuallyTaylor (changes requested):
This all looks really good! Great job!

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+    int8_t getRelativePosition();
+
+private:
+    ///GPIO pin a
+    IO::GPIO& a;
+    ///GPIO pin b
+    IO::GPIO& b;
+
+    ///position can be between = [-range, range]
+    int64_t range;
```
mjmagee991, line 55: I'm beginning to rethink this whole range idea. What if people want a range from -10 to 20? What if they want to increment by 2 each time? What if they want an exponential scale? There's no way to support everything someone might want to do, so I think it might be better to support the minimum and then let the user of the class deal with the details.

Here's a possible implementation: the constructor takes in a max value, an initial position, and a "rollover" boolean. The encoder position then only goes between 0 and the max value, so it can be unsigned. Initial position works as currently implemented. The user of the class can then map encoder values to their values using any implementation they want, so we don't have to support it. The only other parameter I think we might need is rollover, which says if we can roll over from the max value back to 0, which could be useful for selecting from a list of options.

Does this implementation make sense? Are there any features you would add? Do you think it would be better to keep it the way it is?

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+    int8_t calculateChange(int8_t newRelPos);
+
+    /**
+     * sets position to newPosition, capping it if the new value exceeds the range.
+     * @param newPosition the new value for position
+     * @return if position was capped
+     */
+    bool setPosition(int64_t newPosition);
+
+};// namespace EVT::core::DEV
```
mjmagee991, line 109:
```suggestion
};// namespace EVT::core::DEV

```

samples/CMakeLists.txt
```diff
@@ -21,3 +21,4 @@ add_subdirectory(rtc_timer)
 add_subdirectory(thermistor)
 add_subdirectory(timer)
 add_subdirectory(spi)
+add_subdirectory(encoder)
```
mjmagee991, line 24: All files should end with a newline

Also, move this, so it's in alphabetical order

```suggestion
add_subdirectory(encoder)

```

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+    Direction currentDirection;
+    ;///counter for how many updates have occured since the encoder was moved
+    uint8_t noChangeCounter;
+    ///the current relative position of the encoder, in the range: [0,3]
+    int8_t currentRelPos;
+    ///how many updates with no change in the encoder value can occur before the direction resets
+    const static int8_t noChangeCap = 2;
+
+    /**
+     * helper method to convert binary pin values to the relative rotation of the encoder, in the range [0,3]
```
mjmagee991, line 68: I prefer capitalizing the first letter of each comment line, but this is not consistent across our code

samples/encoder/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <cstdio>
+
+namespace IO = EVT::core::IO;
+namespace DEV = EVT::core::DEV;
+namespace time = EVT::core::time;
+
+constexpr IO::Pin A_PIN = IO::Pin::PA_8;
+constexpr IO::Pin B_PIN = IO::Pin::PA_9;
+
+//LCD stuff
```
mjmagee991, line 16: The encoder sample should not have anything related to the charge controller LCD in it

samples/encoder/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+ * encoder B: PA_9
+ *
+ * This is an example designed to work with the charge controller, but please ask someone before flashing the charge controller!
+*/
+constexpr uint8_t deviceCount = 1;
+IO::GPIO* devices[deviceCount];
+
+int main() {
+    //Init platform
+    EVT::core::platform::init();
```
mjmagee991, line 38:
```suggestion
    EVT::core::platform::init();
    
```

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+     */
+    enum Direction {
+        Left = -1,
+        Static = 0,
+        Right = 1
+    };
+
+    /**
+     * Reads and updates the encoder rotation and the absolute position.
+     * @return what position was changed by, should be in this range: [-3,3]
```
mjmagee991, line 30: What use case does this have?

samples/encoder/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+        //ENCODER MUST BE UPDATED EACH LOOP
+        //That is how the rotation is read
+        int8_t change = encoder.update();
+        //Now that the encoder is updated we can read the position of it, which for this example will be between [-124, 124]
+        int64_t position = encoder.getPosition();
+        uint8_t noChangeCounter = encoder.getNoChangeCounter();
+        uint8_t currentRelPos = encoder.getRelativePosition();
+
+        //PRINT VALUES (only enable one at a time)
+        //uart.printf("\r Encoder Change: %d       ", change);
```
mjmagee991, line 58: Since this is a sample, just go ahead and print everything, so it's easy to see the full functionality of the driver

samples/encoder/main.cpp
```diff
@@ -0,0 +1,66 @@
+#include <EVT/dev/LCD.hpp>
```
mjmagee991, line 1:
```suggestion
/**
 * Summary of the purpose of this target (demonstrating how to use Encoder)
 */

#include <EVT/dev/LCD.hpp>
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
+#include <EVT/dev/encoder.hpp>
+
+namespace EVT::core::DEV {
+
+Encoder::Encoder(IO::GPIO& a, IO::GPIO& b, int64_t range, int64_t initialPosition) : a(a), b(b), range(range), position(initialPosition) {
+    //making sure range is positive
+    if (range < 0) {
```
mjmagee991, line 7: This type of error handling is going to be really hard to debug. If you end up restructuring the class the way I described in another comment, this will already get changed, but if not, definitely find another way of handling this

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+    noChangeCounter = 0;
+    currentRelPos = readPinValues();
+}
+
+int64_t Encoder::getPosition() {
+    return position;
+}
+
+int8_t Encoder::update() {
+    //Reads current pin values
```
mjmagee991, line 22: Inline comments should generally be written as commands
```suggestion
    //Read current pin values
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+            break;
+        case 3:
+            setDirection(Left);
+            change = -1;
+            break;
+        default:
+            change = 0;
+        }
+        return change;
+    }
```
mjmagee991, line 58:
```suggestion
    }
    
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+    //Checking if the change is 0, which could either mean the encoder actually moved 0
+    //or the direction is STATIC and the change is 2 or -2, in which case we can't determine which direction
+    //the encoder was turning in. (that case is handled by the default case in the STATIC switch statement)
```
mjmagee991, line 61:
```suggestion
    //the encoder was turning in
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (change == 0) {
+        incrementNoChangeCounter();
+        return change;
+    } else {
+        noChangeCounter = 0;
+    }
+    //If the change is 3 or -3, that's basically it switching from 3 to 0
+    // or 0 to 3, which is really a change of 1 or -1.
+    if (change == 3 || change == -3) {
+        change /= 3;
```
mjmagee991, line 71: Shouldn't the sign flip?

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+    //The only different for non-STATIC directions is that they assume a change of 2
```
mjmagee991, line 73:
```suggestion
    //The only difference for non-STATIC directions is that they assume a change of 2
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
...
+    //find the change in values;
+    int8_t change = calculateChange(newPos);
+    //sets the position to its new value
+    setPosition(position + change);
+    return change;
+}
+
+int8_t Encoder::calculateChange(int8_t newRelPos) {
+    //Gets the change between the positions and then sets currentRelPos to the newRelPos
+    int8_t change = newRelPos - currentRelPos;
```
mjmagee991, line 33: I don't think just subtracting makes sense here because it doesn't have enough information. If the current position is 2, the current direction is right, and it is turned 3 to 1, you're going to get -1 here, which will turn into 1 in the if statements at the bottom of the function

src/dev/encoder.cpp
```diff
@@ -0,0 +1,152 @@
+#include <EVT/dev/encoder.hpp>
```
mjmagee991, line 1: It seems like this class has more functions than it really needs. The idea of breaking down your code does make sense, but there are so many short functions that it is difficult to follow the logic flow

mjmagee991 (changes requested):
Some comments on documentation and class design

include/EVT/dev/encoder.hpp
aclowmclaughlin: All the datasheets just list them as a and b.

include/EVT/dev/encoder.hpp
aclowmclaughlin: If the value gets capped, you can know how much they over-ran the value.

samples/encoder/main.cpp
aclowmclaughlin: The sample updates so fast that if you use newlines it's literally physically not processable. These print statements are single-line overwrites so you can see the values changing in the same place.

src/dev/encoder.cpp
aclowmclaughlin: In any situation in which the encoder is turned 3 ticks between updates, it's impossible to tell between 3 in one direction or -1 in the other. It makes more sense to assume that it was 1 in the other direction, since that is the more likely scenario.

include/EVT/dev/encoder.hpp
aclowmclaughlin: done!

include/EVT/dev/encoder.hpp
aclowmclaughlin: yep! fixed!

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
+#ifndef _EVT_ENCODER_
+#define _EVT_ENCODER_
+
+#include <EVT/io/GPIO.hpp>
+
+namespace EVT::core::DEV {
+class Encoder {
+public:
+    /**
+     * Constructs an Encoder instance. Call update() within the main loop to update encoder values.
```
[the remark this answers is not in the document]
aclowmclaughlin: fixed!

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
+#ifndef _EVT_ENCODER_
+#define _EVT_ENCODER_
+
+#include <EVT/io/GPIO.hpp>
+
+namespace EVT::core::DEV {
```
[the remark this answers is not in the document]
aclowmclaughlin: did it!

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,112 @@
...
+     */
+    int64_t getPosition();
+
+    /**
+     * returns the noChangeCounter value;
+     * @return the noChangeCounter value
+     */
+    uint8_t getNoChangeCounter();
+
+    int8_t getRelativePosition();
```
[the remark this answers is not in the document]
aclowmclaughlin: leftover from debugging, removed it!

include/EVT/dev/encoder.hpp
aclowmclaughlin: Implemented it that way instead, it does make more sense.

samples/encoder/main.cpp
aclowmclaughlin: yeah accidental leftover from debugging, fixed that!

include/EVT/dev/encoder.hpp
aclowmclaughlin: I removed this because it was kind of unnecessary, in retrospect.

src/dev/encoder.cpp
aclowmclaughlin: yep! my bad.

aclowmclaughlin:
Just to let everyone know, I kind of half remade the class to work with interrupts, so many of the changes I made according to your comments are gone now, but I did make them!

src/dev/encoder.cpp
```diff
@@ -61,48 +83,43 @@ void Encoder::update() {
...
-    if (a == 0 && b == 0) {
-        return 0;
-    } else if (a == 1 && b == 0) {
-        return 1;
-    } else if (a == 1 && b == 1) {
-        return 2;
-    } else if (a == 0 && b == 1) {
-        return 3;
-    }
+    if (a == 0 && b == 0) {return 0;}
```
ActuallyTaylor: I may be wrong but I think our formatting for most repositories follows along the idea that if statements should be expanded out. If not expanded at least a heavy use of whitespace. But I my be wrong

include/EVT/dev/encoder.hpp
ActuallyTaylor: Okay sounds good!

src/dev/encoder.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+int8_t Encoder::readPinValues() {
+    bool aPos = (bool) a.readPin();
+    bool bPos = (bool) b.readPin();
+    int8_t newPos = convertPinValuesToPosition(aPos, bPos);
+    return newPos;
+}
+
+int8_t Encoder::convertPinValuesToPosition(bool a, bool b) {
```
ActuallyTaylor, line 89: Is this function used anywhere else then on line 85? If not could it just be moved into that function instead of an entire standalone function?

ActuallyTaylor (approved):
It all looks good for the most part. 

I have a few questions about the general style of comments but I think that is better suited for a team discussion on style we want to go with going forward.

ActuallyTaylor (approved):
It all looks good for the most part. 

I have a few questions about the general style of comments but I think that is better suited for a team discussion on style we want to go with going forward.

aclowmclaughlin:
Okay so it is ACTUALLY working. This should be the final build once you all review it and I address all your comments.

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * update() must be called within the main loop to update encoder position.
+ */
+class Encoder {
+public:
+    /**
+     * Constructs an Encoder instance.
+     *
+     * @param[in] a a pin of the encoder
+     * @param[in] b b pin of the encoder
+     * @param range range of the encoder positions
```
mjmagee991, line 25: Here and in the functions below
```suggestion
     * @param[in] range range of the encoder positions
```

include/EVT/dev/encoder.hpp
mjmagee991, line 1: This and the `.cpp` file should be _**E**_ ncdoer, note encoder

samples/encoder/main.cpp
```diff
@@ -0,0 +1,71 @@
+/**
+ * This is an example designed to work with the charge controller,
```
mjmagee991, line 2: An example shouldn't be targeted at specific hardware. Here, obviously, you are using specific hardware for this test, but the explanation should just talk about what you're trying to test and what success is supposed to look like

samples/encoder/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+namespace IO = EVT::core::IO;
+namespace DEV = EVT::core::DEV;
+namespace time = EVT::core::time;
+namespace log = EVT::core::log;
+
+constexpr IO::Pin A_PIN = IO::Pin::PA_8;
+constexpr IO::Pin B_PIN = IO::Pin::PA_9;
+
+constexpr uint8_t deviceCount = 1;
+IO::GPIO* devices[deviceCount];
```
mjmagee991, line 22: What is this?

samples/encoder/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+//Necessary static wrappers for the encoder to handle the pin interrupts
+DEV::Encoder* encoderWrapper;
```
mjmagee991, line 25: I would still like you to get rid of the global stuff here. To do this, you're going to have to modify `GPIO.hpp`, `GPIOf3xx.hpp`, and `GPIOf3xx.cpp`. You're going to want to add a `void*` to the accepted interrupt handler, and you're going to want to store the private data as well. Line 105 of `GPIO.hpp` should become the following: 
`virtual void registerIRQ(TriggerEdge edge, void (*irqHandler)(GPIO* pin, void* priv), void* priv) = 0;`

In `GPIOf3xx.cpp`, you're then going to have to modify `registerIRQ` and HAL_GPIO_EXTI_Callback. You're probably also going to have to add an array after `INTERRUPT_GPIOS` (line 13) that looks something like this:
`void* INTERRUPT_DATA[16] = {nullptr};`

Let me know if this doesn't make sense.

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+     * @return the current position of the encoder, between 0 and range, inclusive
+     */
+    uint64_t getPosition() const;
+
+    /**
+     * Handles the interrupt generated by pin a. If two interrupts occur too soon after one another the second is ignored.
+     *
+     * @param pin pin that cause the interrupt
+     */
+    void aInterruptHandler(IO::GPIO* pin);
```
mjmagee991, line 47: You don't need to pass the GPIO in here. You don't use it

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+    /** position can be between = [0, range] */
+    uint64_t range;
+    /** the current absolute position of the encoder */
+    uint64_t position;
+    /** if the position rolls over when it reaches range or 0 */
+    bool rollOver;
+    /** counter for how many updates have occurred since the encoder was moved */
+    int8_t currentRelPos;
+
+    ///INTERRUPT RELATED VARIABLES
```
mjmagee991, line 71: Triple slashes triggers Doxygen, which we don't want here
```suggestion
    //INTERRUPT RELATED VARIABLES
```

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+     */
+    void bInterruptHandler(IO::GPIO* pin);
+
+private:
+    /** GPIO pin a */
+    IO::GPIO& a;
+    /** GPIO pin b */
+    IO::GPIO& b;
+
+    /** position can be between = [0, range] */
```
mjmagee991, line 62:
```suggestion
    /** position can be in [0, range] */
```

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+    /** if the position rolls over when it reaches range or 0 */
+    bool rollOver;
+    /** counter for how many updates have occurred since the encoder was moved */
+    int8_t currentRelPos;
+
+    ///INTERRUPT RELATED VARIABLES
+    /** the change since the last update was called, updated in the interrupt handler */
+    int64_t interruptChange;
+    /** the most recent time in milliseconds since the last interrupt on pin A */
+    uint32_t recentAInterruptTime;
```
mjmagee991, line 75: This is a nitpick, but would this make more sense?
```suggestion
    uint32_t lastAInterruptTime;
```

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+     * reads pin values and converts them to the relative position
+     *
+     * @return relative position of the encoder
+     */
+    int8_t readPinValues();
+
+    /**
+     * changes position by change, capping or rolling over depending on rollOver
+     *
+     * @param newPosition the new value for position
```
mjmagee991, line 100: This is no longer accurate

src/dev/encoder.cpp
mjmagee991: Yeah, this logic is only really going to be useful in `readPinValues`, so I think it makes more sense to be in there

samples/encoder/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    //setting the wrapper to the encoder
+    encoderWrapper = &encoder;
+
+    //Creating the interrupts
+    pinA.registerIRQ(IO::GPIO::TriggerEdge::RISING_FALLING, ainterruptHandlerWrapper);
+    pinB.registerIRQ(IO::GPIO::TriggerEdge::RISING_FALLING, binterruptHandlerWrapper);
+
+    while (1) {
+        //ENCODER MUST BE UPDATED EACH LOOP
```
mjmagee991, line 58: What if we just update when we `getPosition()`?

samples/encoder/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+        encoder.update();
+        //Now that the encoder is updated we can read the position of it, which for this example will in the range [0, 18]
+        uint64_t position = encoder.getPosition();
+
+        //PRINT VALUES (only enable one at a time)
+        //uart.printf("\r Encoder Change: %d       ", change);
+        uart.printf("\rPosition: %d         ", position);
+
+        //The wait simulates a loop that is doing other processing, because that will affect how often the output is read
+        time::wait(100);
```
mjmagee991, line 69: Crank this up to show how powerful your interrupt-based solution is. The sample should push the driver to its limits to show its strengths and weaknesses.
```suggestion
        time::wait(300);
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,141 @@
...
+namespace EVT::core::DEV {
+
+Encoder::Encoder(IO::GPIO& a, IO::GPIO& b, uint64_t range, uint64_t initialPosition, bool rollOver)
+    : a(a), b(b), range(range), position(initialPosition), rollOver(rollOver) {
+    if (position > range) {
+        position = range;
+    }
+    //setting instance variables
+    currentRelPos = readPinValues();
+    interruptChange = 0;
```
mjmagee991, line 12: This and the two below can be set by default in the `.hpp` instead of here. Putting them there just makes this method harder to mess up in the future

src/dev/encoder.cpp
```diff
@@ -0,0 +1,141 @@
...
+        currentRelPos = 3;
+        break;
+    case 3:
+        change = -1;
+        currentRelPos = 2;
+        break;
+    }
+    interruptChange += change;
+    recentAInterruptTime = time::millis();
+    //time::wait(200);
```
mjmagee991, line 43:
```suggestion
```

src/dev/encoder.cpp
```diff
@@ -0,0 +1,141 @@
...
+    } else if (a == 1 && b == 0) {
+        return 1;
+    } else if (a == 1 && b == 1) {
+        return 2;
+    } else if (a == 0 && b == 1) {
+        return 3;
+    }
+}
+
+bool Encoder::changePosition(int64_t change) {
```
mjmagee991, line 115: This function will break if you are using ranges, positions, etc that are greater than 2^63. While that is unlikely, you should rework this, so that it doesn't have that issue

include/EVT/dev/encoder.hpp
```diff
@@ -0,0 +1,108 @@
...
+     * @return if position tried to go outside the range
```
mjmagee991, line 101: Why are you returning this if you don't use it?

mjmagee991 (changes requested):
Some more implementation cleanup

samples/encoder/main.cpp
aclowmclaughlin: leftover from LCD display, removed it!

include/EVT/dev/encoder.hpp
aclowmclaughlin: It's necessary because registerIRQ requires you to to pass in a void method that takes in a GPIO pointer. If it does not have a GPIO pointer it does not work.

include/EVT/dev/encoder.hpp
aclowmclaughlin: It actually is completely accurate. It is used in the update method to update the position.

samples/encoder/main.cpp
aclowmclaughlin: Great idea, done!

include/EVT/dev/encoder.hpp
aclowmclaughlin: I figured in a situation where they're not rolling over it could be good to know if they're hitting the cap.

include/EVT/dev/encoder.hpp
aclowmclaughlin: done!

include/EVT/dev/encoder.hpp
aclowmclaughlin: renamed them!

include/EVT/dev/encoder.hpp
aclowmclaughlin: Nevermind I see what you mean, fixed! My bad!

src/dev/encoder.cpp
aclowmclaughlin: fixed!

src/dev/encoder.cpp
aclowmclaughlin: Moved it!

mjmagee991 (changes requested):
We just merged the F4xx stuff into master, so you're going to have to back merge (merge the new version of master into your branch) and then update GPIOF4xx to match your changes to GPIOF3xx. Hopefully, this will be pretty straightforward

ActuallyTaylor (approved):
Looks good! I like the use of the logger.

mjmagee991 (approved):
Nice work!

PR #91, opened 2023-11-18
This Encoder is Driving me crazy (Dynamically!)

aclowmclaughlin:
Added the ability to dynamically change the encoder range and position. This is important for the general use case of an encoder being used as an input device to a UI, because different UI menus may have different numbers of options.
I should have implemented this in the original pull request, but I forgor.

Also removed an hpp method that wasn't implemented anymore.

### DIFF
--- a/include/EVT/dev/Encoder.hpp
+++ b/include/EVT/dev/Encoder.hpp
@@ -28,16 +28,19 @@ public:
     Encoder(IO::GPIO& a, IO::GPIO& b, uint32_t range, uint32_t initialPosition, bool rollOver);
 
     /**
-     * Reads and updates the encoder position.
-     */
-    void update();
-
-    /**
      * Returns the current absolute position
      *
      * @return the current position of the encoder, between 0 and range, inclusive
      */
     uint64_t getPosition();
+
+    /**
+     * Sets the range and position of the encoder
+     *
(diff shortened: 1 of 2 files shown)


include/EVT/dev/Encoder.hpp
```diff
@@ -27,18 +27,21 @@ class Encoder {
...
      */
     uint64_t getPosition();
 
+    /**
+     * Sets the range and position of the encoder
+     *
+     * @param newRange the new range for the encoder positions
+     * @param newPosition a new position for the encoder (defaults to 0)
+     */
+    void setRangeAndPosition(uint32_t newRange, uint32_t newPosition = 0);
```
mjmagee991, line 43: I think it would make more sense to not have a default value for this, but if you think there's good reason for it, it's probably okay to leave

mjmagee991 (approved):
One suggestion, but otherwise looks good

include/EVT/dev/Encoder.hpp
aclowmclaughlin: I just figure in most (if not all) circumstances you'd want the position to reset to 0, but I didn't want to lose the ability to specify if necessary.

ActuallyTaylor (approved):

PR #94, opened 2024-01-27
I am encoding buttons to be high 🔨 🐛 

aclowmclaughlin:
Bugfix for the encoder class (was modding by range instead of range + 1).
Also fixed some return values for functions, the getPosition() should have been a uint32_t all along I'm just dumb as hell.

Also added the ability to define what GPIO state (high or low) is considered "pressed" for button.hpp.


tmb5932 (approved):
This shit 🔥

ActuallyTaylor (approved):
looks chill

PR #100, opened 2024-03-30
F4xx SPI Support. Was too easy. didn't get enough time to hate myself and think of a pun

tmb5932:
Support for STM32F446 SPI

### DIFF
--- a/include/EVT/io/pin.hpp
+++ b/include/EVT/io/pin.hpp
@@ -128,10 +128,10 @@ enum class Pin {
 #endif
 
 #ifdef STM32F4xx
-    SPI_SCK = PA_5,     //D13
-    SPI_MISO = PA_6,    //D12
-    SPI_MOSI = PA_7,    //D11
-    SPI_CS = PB_6,      //D10
+    SPI_SCK = PA_9,     // D8 todo give the Dxx pin label for these
+    SPI_MISO = PC_2,    //
+    SPI_MOSI = PC_3,    //
+    SPI_CS = PB_4,      // D5
     I2C_SDA = PB_9,
     I2C_SCL = PB_8,
 #endif
--- a/include/EVT/io/pin.hpp
+++ b/include/EVT/io/pin.hpp
@@ -111,23 +111,27 @@ enum class Pin {
 #endif
 
 #ifdef STM32F302x8
-    SPI_SCK = PB_13, //D13
-    SPI_MISO = PB_14,//D12
-    SPI_MOSI = PB_15,//D11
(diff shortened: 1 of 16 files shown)


ol2764RIT (approved):
good job travis, cooking up bangers left and right bb

include/EVT/io/pin.hpp
```diff
@@ -128,6 +128,10 @@ enum class Pin {
 #endif
 
 #ifdef STM32F4xx
+    SPI_SCK = PA_9,// D8
```
mjmagee991, line 131: Why did you label two of them with their D values but not the other two?

include/EVT/manager.hpp
```diff
@@ -52,7 +53,7 @@
     //    #include <EVT/io/platform/f4xx/PWMf4xx.hpp>
```
mjmagee991, line 53: Delete PWM and I2C from here because they've already been implemented

src/io/platform/f4xx/SPIf4xx.cpp
```diff
@@ -0,0 +1,281 @@
...
+SPI::SPIStatus SPIf4xx::read(uint8_t* bytes, uint8_t length) {
+    HAL_StatusTypeDef halStatus = HAL_SPI_Receive(&halSPI, bytes, length, EVT_SPI_TIMEOUT);
+
+    return halToSPIStatus(halStatus);
+}
+
+/**
+ * Converts HAL Status into EVT Core's own SPIStatus enum
+ *
+ * @param halStatus
```
mjmagee991, line 269:
```suggestion
 * @param[in] halStatus HAL-defined status enum value
```

mjmagee991 (approved):
A couple details to clean up but looks pretty good

include/EVT/io/pin.hpp
tmb5932: well... i meant to talk to you about that but clearly forgot. The other 2 don't have a D label. Only SPI1 has the capabilities to be only on D's, but SPI1's necessary pins are being used by GPIO by default, so it isn't usable normally. I'll talk to you about it at EVT tonight.

aclowmclaughlin (changes requested):
Looks mostly good.

include/EVT/io/pin.hpp
tmb5932: SPI1 is now the default SPI being used. Going to test it on thurs but everything should be in order.

aclowmclaughlin (approved):
Sure looks like code! (Good job)

PR #103, opened 2024-07-06
Making Better Options

mjh9585:
Added a memory summary to the cmake build configuration to print memory usage for each target. Also added an option to force color output when using the command line with ninja. 

- Added memory usage summary to builds
- Added FORCE_COLORED_OUTPUT option to always colorize build output

### DIFF
--- a/cmake/evt-core_compiler.cmake
+++ b/cmake/evt-core_compiler.cmake
@@ -33,10 +33,10 @@ else()
 endif()
 message(STATUS "${TARGET_DEV} targeted")
 
-option (FORCE_COLORED_OUTPUT "Always produce ANSI-colored output (GNU/Clang only)." FALSE)
-if (${FORCE_COLORED_OUTPUT})
-    add_compile_options (-fdiagnostics-color=always)
-endif ()
+option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored output (GNU/Clang only)." OFF)
+if(FORCE_COLORED_OUTPUT)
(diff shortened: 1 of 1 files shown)


cmake/evt-core_compiler.cmake
```diff
@@ -32,6 +33,10 @@ else()
 endif()
 message(STATUS "${TARGET_DEV} targeted")
 
+option (FORCE_COLORED_OUTPUT "Always produce ANSI-colored output (GNU/Clang only)." OFF)
```
mjmagee991, line 36: I would argue that we would want this on by default, but I don't really care

cmake/evt-core_compiler.cmake
```diff
@@ -32,6 +33,10 @@ else()
...
+if (FORCE_COLORED_OUTPUT)
```
mjmagee991, line 37: You put spaces after all your keywords, but no other lines in this file do that. I would prefer that you conform.

mjmagee991 (approved):

cmake/evt-core_compiler.cmake
mjh9585: If gcc has access to the console then it does this by default, however a build tool like ninja pipes the output to the program so gcc disables it. This just adds the flag to force colorizing. Not everyone (for some reason) like it and something like github actions shouldn't have to deal with that.

cmake/evt-core_compiler.cmake
mjh9585: work habit, it is invading.

PR #104, opened 2024-07-20
FWT-52 RTOS'n and turning in my dreams ༼ つ ◕_◕ ༽つ💨🛏

DiegoLHendrix:
Adding Threadx into EVT-core in order to be able to use an RTOS. 💯💯💯
- Added Threadx submodule. 🥶
- Moved assembly file to RTOS platform.
- Edited Cmake files to be able to compile Threadx and the demos.👅


src/RTOS/f4xx/tx_initialize_low_level.s
mjmagee991, line 1: The `RTOS` folder should be renamed `rtos`, so it matches the other folder names. Also, where did this file come from? Why did we need to add a random assembly file?

CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ file(STRINGS version.txt EVT_VERSION)
 
 project(EVT
     VERSION ${EVT_VERSION}
-    LANGUAGES CXX C
+    LANGUAGES CXX C ASM
```
mjmagee991, line 24: Why do we need to compile assembly like this, while we don't do that for the other assembly files we have?

CMakeLists.txt
```diff
@@ -109,6 +111,17 @@ install(EXPORT canopen-stack-config DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ca
...
 target_link_libraries(${PROJECT_NAME} PUBLIC canopen-stack)
 
+set(THREADX_ARCH "cortex_m4")
+set(THREADX_TOOLCHAIN "gnu")
+set(MCPU_FLAGS "-mthumb -mcpu=cortex-m4")
+set(VFP_FLAGS "")
+set(SPEC_FLAGS "--specs=nosys.specs")
+
+add_subdirectory(libs/threadx)
+#install(TARGETS threadx EXPORT threadx-config)
```
mjmagee991, line 121: If these are not necessary, they should be deleted, not left commented out

cmake/evt-core_compiler.cmake
```diff
@@ -102,4 +105,4 @@ endif()
 
 set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} -mfloat-abi=hard \
                             -specs=nano.specs -specs=nosys.specs \
-                            -lc -lm -lnosys -Wl,--gc-section")
+                            -lc -lm -lnosys -Wl,--gc-section -Wl,--print-memory-usage")
```
mjmagee991, line 108: As discussed in this issue: https://rit-evt.atlassian.net/browse/FWT-64, the `gc-section` flag is important for reducing linking of unnecessary objects. Why have you removed it here?

cmake/evt-core_build.cmake
```diff
@@ -26,4 +26,5 @@ macro(make_exe proj_name sources)
 
     # Link the EVT-core library
     target_link_libraries(${proj_name} PUBLIC EVT)
+    target_link_libraries(${proj_name} PUBLIC threadx)
```
mjmagee991, line 29: Why do we need this but not an equivalent call for the CANopen stack

samples/RTOS/Basic-Test/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+include(../../../cmake/evt-core_build.cmake)
+
+set(SAMPLE_SOURCES main.cpp)
```
mjmagee991, line 3: Why did you create this CMake variable just to use it on the next line?

samples/RTOS/Dynamic-Threads/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+include(../../../cmake/evt-core_build.cmake)
+
+set(SAMPLE_SOURCES main.cpp)
```
mjmagee991, line 3: Why did you create this variable?

samples/RTOS/Basic-Test/CMakeLists.txt
mjmagee991, line 1: The `Basic-Test` folder should be renamed to something in lower_snake_case. Also, could you give it a more descriptive name than "basic test"?

samples/RTOS/Basic-Test/main.cpp
```diff
@@ -0,0 +1,285 @@
+//
```
mjmagee991, line 1: Delete this automatically generated comment and add a comment that follows our normal standard

samples/RTOS/Basic-Test/main.cpp
mjmagee991, line 1: A bunch of this file doesn't follow our standards, but I'll let it be for now because it's all going to have to get rewritten when we add the C++ wrapper layer

samples/RTOS/Dynamic-Threads/main.cpp
```diff
@@ -0,0 +1,238 @@
+#include "tx_api.h"
```
mjmagee991, line 1: Add a comment at the top of this file explaining what this sample is for

samples/RTOS/Dynamic-Threads/main.cpp
mjmagee991, line 1: This sample also has a bunch of issues, which I'll ignore until after the C++ wrapper layer is in

mjmagee991 (changes requested):
A couple little changes and a bunch of questions

src/RTOS/f4xx/tx_initialize_low_level.s
mjh9585: This is the funny assembly file I mentioned about 2 weeks ago. The file is from cube and it initialized a bunch of specific hardware registers needed for ThreadX to run. Looking at it, the file is mosly configuring systick and the NVIC with specific settings.

cmake/evt-core_compiler.cmake
mjh9585: Where do you see it being removed? I see `-Wl,--gc-section -Wl,--print-memory-usage` which adds another flag, nothing is removed

cmake/evt-core_build.cmake
mjh9585: This is probably what we needed to do with the CANOpen stack instead of all the modifications we made to the original library.

src/RTOS/f4xx/tx_initialize_low_level.s
mjmagee991: I believe that it's important, but I was asking why the startup assembly files like startup_stm32f302x8.s don't need the compiler set up like this. Could we pull anything from HALf3/CMakeLists.txt? Or is this the best solution?

cmake/evt-core_compiler.cmake
mjmagee991: Well, I'm either blind or stupid. Although, I believe this could be simplified to 
```suggestion
                            -lc -lm -lnosys -Wl,--gc-section,--print-memory-usage")
```

cmake/evt-core_build.cmake
mjmagee991: Do you think it's worth undoing what we did and switching over to this method, so it's easier to work with next time we update the CANopen stack?

mjmagee991: Also, @DiegoLHendrix, you should update the comment on line 27

mjmagee991 (changes requested):

CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ elseif(COMPDEFS MATCHES "(.*)STM32F4xx(.*)")
     src/io/platform/f4xx/SPIf4xx.cpp
     src/dev/platform/f4xx/IWDGf4xx.cpp
     src/dev/platform/f4xx/RTCf4xx.cpp)
+
```
mjmagee991, line 77: You can leave this if you don't feel like fixing it, but I really prefer we don't make changes that aren't necessary
```suggestion
```

samples/rtos/dynamic-threads/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include(../../../cmake/evt-core_build.cmake)
```
mjmagee991, line 1:
```suggestion
include(${CMAKE_CURRENT_SOURCE_DIR}/../../../cmake/evt-core_build.cmake)
```

samples/rtos/threadx-demo/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include(../../../cmake/evt-core_build.cmake)
```
mjmagee991, line 1:
```suggestion
include(${CMAKE_CURRENT_SOURCE_DIR}/../../../cmake/evt-core_build.cmake)
```

mjmagee991 (approved):
Great work!

samples/rtos/dynamic-threads/main.cpp
```diff
@@ -0,0 +1,241 @@
...
+ULONG generator_sum = 0;
+
+///Function declarations
+void generator_entry(ULONG thread_input) {
+    // Setup UART
+    IO::UART& uart = IO::getUART<IO::Pin::UART_TX, IO::Pin::UART_RX>(9600);
+
+    srand(77);
+    ULONG queue_status;
+    ULONG semaphore_status;
```
mjh9585, line 118:
```suggestion
```
You should remove unused variables and avoid having compiler warnings.

mjh9585 (approved):
Looks good. The samples could be cleaned up some more, but for now they are fine until we can get the new C++ conversion layer in.

ol2764RIT:
peak diego

PR #106, opened 2024-09-13
FWT-19 F4xx - Timer? I Barely Know Her!

tmb5932:
Implemented F4xx support for Timer, and tested that all the timers work. 
Fixed some issues that existed involving F3xx while I was at it :).


src/dev/platform/f4xx/Timerf4xx.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+/**
+ * Returns the corresponding index for the given timer device as stored in the
+ * timerInterruptHandlers array
+ * @param htim the TIM_TypeDef pointer to the timer device
+ * @return an uint8_t between 0-9 corresponding to an element of timerInterruptHandlers
+ */
+uint8_t getTimerInterruptIndex(TIM_TypeDef* peripheral) {
```
ol2764RIT, line 159: Turn this into a switch statement in my opinion, too many else if's

include/EVT/dev/platform/f4xx/Timerf4xx.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+    void reloadTimer() override;
+
+    void setPeriod(uint32_t clockPeriod) override;
+
+private:
+    // Pointer to the halTimer struct stored in the global array in Timerf4xx.cpp
+    TIM_HandleTypeDef* halTimer;
+
+    // Timer clock period
```
ol2764RIT, line 43: For clockPeriod being stored as a uint32 it's probably fine, but to nit pick you can add a check if the timer runs on too long and you should throw a error

src/dev/platform/f4xx/Timerf4xx.cpp
```diff
@@ -0,0 +1,252 @@
...
+void Timerf4xx::startTimer() {
+    stopTimer();// Stop timer in case it was already running
+
+    auto htim = this->halTimer;
+    // Clear the interrupt flag so interrupt doesn't trigger immediately
+    __HAL_TIM_CLEAR_IT(htim, TIM_IT_UPDATE);
+    HAL_TIM_Base_Start_IT(htim);
+}
+
+void Timerf4xx::reloadTimer() {
```
ol2764RIT, line 244: In reloadTimer(), clearing the counter register with this->halTimer->Instance->CNT &= ~(0xFFFFFFFF) can be simplified to just setting the counter to 0: this->halTimer->Instance->CNT = 0;.

src/dev/platform/f4xx/Timerf4xx.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    // Timers 9-14 are NOT master mode compatible, so waste of time to go through config
+    if (getTimerInterruptIndex(timerPeripheral) < static_cast<uint8_t>(timerInterruptIndex::TIM9_IDX)) {
+        masterConfig.MasterOutputTrigger = TIM_TRGO_RESET;
+        masterConfig.MasterSlaveMode = TIM_MASTERSLAVEMODE_DISABLE;
+        HAL_TIMEx_MasterConfigSynchronization(&htim, &masterConfig);
+    }
+}
+
+void Timerf4xx::startTimer(void (*irqHandler)(void* htim)) {
```
ol2764RIT, line 224: In start time you run stop timer. Lowkey, it might be more efficient to first check if the timer is actually running before stopping it to avoid redundant operations.

ol2764RIT (changes requested):
travis with the peak f4 timer class

src/dev/platform/f4xx/Timerf4xx.cpp
tmb5932: The peripheral that is being if'ed on is a memory address for the timer, which as far as i know cannot be used in a switch (i think)

mjmagee991 (changes requested):
Minor details

src/platform/f4xx/stm32f446xx.cpp
```diff
@@ -42,7 +42,7 @@ void stm32f4xx_init() {
     */
     RCC_ClkInitStruct.ClockType = RCC_CLOCKTYPE_HCLK | RCC_CLOCKTYPE_SYSCLK
         | RCC_CLOCKTYPE_PCLK1 | RCC_CLOCKTYPE_PCLK2;
-    RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_PLLCLK;// SYSCLK at 8 MHz (HSI CLK)
+    RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_PLLCLK;// SYSCLK at 50 MHz (HSI CLK)
```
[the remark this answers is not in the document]
tmb5932: it currently is not used anywhere in f4, and in f3 it is only used in the timer cpp.  Reason it isn't used in f4 timer is cause Heller said to just use SystemCoreClock instead of this variable.

src/dev/platform/f4xx/Timerf4xx.cpp
tmb5932: Are you saying with `enum` instead of `enum class` I don't need to do the static cast to uint8, or are you saying something else is possible with `enum` instead of `enum class`?

src/platform/f4xx/stm32f446xx.cpp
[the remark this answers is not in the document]
mjmagee991: Just because the variable isn't used doesn't mean it won't break things. This is affecting the hardware configuration, so it may break things that previously worked in hardware. Any thoughts @mjh9585?

src/dev/platform/f4xx/Timerf4xx.cpp
mjmagee991: > Are you saying with `enum` instead of `enum class` I don't need to do the static cast to uint8

Yes, this is exactly what I'm saying. You can pass an `enum` into a function asking for `uint8_t`

src/platform/f4xx/stm32f446xx.cpp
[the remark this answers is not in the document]
tmb5932: Just let me know what yall think. I'll revert or w/e Matt and Matt suggest 🐸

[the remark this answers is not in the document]
mjh9585: I am confused what we are referring to since the code referenced is just changing a comment.  If you are referring to CLK_SPEED in stm32fxx.hpp then that is fine as it was already wrong since the chip is running at 50MHz, not 8MHz. Additionally for the timer calculations, Travis is using SystemCoreClock which is calculated when the system clock gets update and is a much better source then a hard coded value.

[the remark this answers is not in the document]
mjmagee991: In my head, I was assuming that he actually changed something other than the comment to change the frequency, but if this is just updating documentation to be more accurate, I'm fine with it

src/dev/platform/f3xx/Timerf3xx.cpp
```diff
@@ -142,7 +142,10 @@ void Timerf3xx::initTimer(TIM_TypeDef* timerPeripheral, uint32_t clockPeriod) {
 
 void Timerf3xx::startTimer(void (*irqHandler)(void* htim)) {
     TIM_TypeDef* timerPeripheral = this->halTimer->Instance;
-    stopTimer();
+    // If timer is not waiting to start, stop it
+    if (halTimer->State != HAL_TIM_STATE_READY)
```
mjmagee991, line 146: Always use braces even if you don't need them. This is liable to cause accidents later
```suggestion
    if (halTimer->State != HAL_TIM_STATE_READY) {
```

mjmagee991 (approved):
Nice job

ol2764RIT (approved):
ترافيس يطبخ!
Type things

PR #107, opened 2024-09-15
FWT-17 Clocking in at over 32MHz

mjh9585:
Fixes the F302 and F334 system clock to run at 32MHz instead of the current 8MHz. Also makes the corresponding hardware drivers aware of the system clock when determining prescalers.

- Changed main system clock source to the PLL CLK at 32MHz
- Corrected ADC clock to support the added speed
- Calculated PWM and Timer prescalers using the System clock speed instead
- Fix SPI timing
- Fixed CAN prescaler


src/core/io/platform/f3xx/SPIf3xx.cpp
```diff
@@ -241,23 +241,18 @@ void SPIf3xx::configureSPI(uint32_t baudRate, SPIMode mode, bool firstBitMSB) {
...
-        halSPI.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_128;
-    } else {
-        halSPI.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256;
+    uint32_t prescaler = (HAL_RCC_GetHCLKFreq() / baudRate);
+    // Convert the prescaler number to the bit value incrementally (log2)
+    halSPI.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_2;
+    while (prescaler >>= 1) {
+        halSPI.Init.BaudRatePrescaler++;
+    }
+    // Max value is a prescaler of 256 or 0x07+1.
```
mjmagee991, line 250: > 0x07+1
What?

src/core/io/platform/f3xx/SPIf3xx.cpp
```diff
@@ -241,23 +241,18 @@ void SPIf3xx::configureSPI(uint32_t baudRate, SPIMode mode, bool firstBitMSB) {
...
+    if (halSPI.Init.BaudRatePrescaler > 0x8) {
+        halSPI.Init.BaudRatePrescaler = 0x8;
     }
+    // Shift left 3 to correct bit position.
+    halSPI.Init.BaudRatePrescaler = (halSPI.Init.BaudRatePrescaler - 1) << 3;
```
mjmagee991, line 255: Can you add comments to make it clearer why you're doing this?

mjmagee991 (approved):
Question on one part, but looks good otherwise

src/core/platform/f3xx/stm32f302x8.cpp
```diff
@@ -20,13 +20,13 @@ void stm32f3xx_init() {
     RCC_OscInitStruct.HSICalibrationValue = RCC_HSICALIBRATION_DEFAULT;
     RCC_OscInitStruct.PLL.PLLState        = RCC_PLL_ON;
     RCC_OscInitStruct.PLL.PLLSource       = RCC_PLLSOURCE_HSI;
-    RCC_OscInitStruct.PLL.PLLMUL          = RCC_PLL_MUL4; // PLL CLK at 8 * 4 = 32 MHz
+    RCC_OscInitStruct.PLL.PLLMUL          = RCC_PLL_MUL8; // PLL CLK at 8 / 2 * 8 = 32 MHz
     HAL_RCC_OscConfig(&RCC_OscInitStruct);
 
     /** Initializes the CPU, AHB and APB buses clocks
      */
```
tmb5932, line 27: Unimportant change but just make this a 1 line comment instead of having the `*/` on the next line, no?

tmb5932 (approved):
God, IDK who this Heller guy is but hes pretty good at this EVT Firmware stuff...

src/core/io/platform/f3xx/SPIf3xx.cpp
mjh9585: A prescaler of 256 is 0x7 but due to the loop the value for the calculated prescaler will be one more (+1) this is then subtracted before shifting in the next step to bits 5:3 for the BR field.

mjmagee991: Some explanation like that would do nicely in a comment around here

PR #108, opened 2024-09-28
FWT-45 adding Multiple Actually Dangerous Commits (Multi ADC Support)

tmb5932:
Super cool really cool working multi adc for F4xx. Also made F4xx single adc work properly, as apparently it was broken before 👎


include/core/io/platform/f3xx/ADCf3xx.hpp
```diff
@@ -8,14 +8,18 @@
...
+
 class ADCf3xx : public ADC {
 public:
     /**
      * Setup the given pin for ADC usage
      *
      * @param[in] pin The pin to setup for ADC
      */
-    ADCf3xx(Pin pin);
+    ADCf3xx(Pin pin, ADCPeriph adcPeriph);
```
mjh9585, line 22: Default value?

samples/adc/single_adc.cpp
```diff
@@ -21,7 +21,7 @@ int main() {
 
     time::wait(500);
 
-    io::ADC& adc0 = io::getADC<io::Pin::PA_0>();
+    io::ADC& adc0 = io::getADC<io::Pin::PA_0, io::ADCPeriph::THREE>();
```
mjh9585, line 24: This sample is also used by the F3xx so this should be left as the default.

src/core/io/ADC.cpp
```diff
@@ -2,8 +2,9 @@
 
 namespace core::io {
 
-ADC::ADC(Pin pin) {
-    this->pin = pin;
+ADC::ADC(Pin pin, ADCPeriph adcPeriph) {
+    this->pin       = pin;
+    this->adcPeriph = adcPeriph;
```
mjh9585, line 7: This could be an initializer list.

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +98,189 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+        break;
+    case ADC2_SLOT:
+        halADC[adcNum].Instance = ADC2;
+        break;
+    case ADC3_SLOT:
+        halADC[adcNum].Instance = ADC3;
+        break;
+    default:
+        return; // Should never get here
+    }
```
mjh9585, line 116: This could be done when you initialize halADC since the Instance is just a pointer to the ADC memory address

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +98,189 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+        channel = 0;
         break; // Should never get here
     }
 
+    // This checks if the pin being used supports the ADC being used
+    if (checkSupport(adcPeriph, channel)) {
+        // Masks channel back to proper value (Zero's out ADC information bits)
+        adcChannel.Channel = channel & 0x1F;
+    } else {
+        // Causes HARD FAULT if pin does not support the ADC peripheral being used
```
mjh9585, line 248:
```suggestion
        // Causes HARD FAULT if pin does not support the ADC peripheral being used. THIS IS INTENTIONAL!
```

include/core/io/platform/f4xx/ADCf4xx.hpp
```diff
@@ -34,14 +42,26 @@ class ADCf4xx : public ADC {
...
     /// Static list of all channels supported by the ADC
-    static Pin channels[MAX_CHANNELS];
-    /// Buffer for DMA where each spot represents the value read in from a
-    /// channel
-    static uint16_t buffer[MAX_CHANNELS];
-    static DMA_HandleTypeDef halDMA;
+    static Pin channels[NUM_ADCS][MAX_CHANNELS];
+    /// Buffer for DMA where each spot represents the value read in from a channel
+    static uint16_t buffer[NUM_ADCS][MAX_CHANNELS];
+    static DMA_HandleTypeDef halDMA[NUM_ADCS];
```
mjh9585, line 50: I would suggest making a struct that holds all of this data then make an array of 3 structs for your ADCs. This way you index the array once at the beginning of a method using a pointer and the rest of the code focus on that single ADC.

mjh9585 (changes requested):
I don't see anything obvious that is causing your issue, but maybe using a struct for each ADC will better distinguish between the ADCs and fix your issue in the process.

include/core/io/platform/f3xx/ADCf3xx.hpp
tmb5932: Default value is set in the template thingy in manager for adc

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
 #include <HALf4/stm32f4xx.h>
 #include <HALf4/stm32f4xx_hal_adc.h>
 #include <core/io/pin.hpp>
 #include <core/io/platform/f4xx/ADCf4xx.hpp>
 #include <core/io/platform/f4xx/GPIOf4xx.hpp>
+#include <core/platform/f4xx/stm32f4xx.hpp>
+#include <core/utils/log.hpp>
+
+namespace {
+/// This is made as a global variable so that it is accessible in the interrupt.
```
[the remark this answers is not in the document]
mjmagee991: Also, why put these here instead of in the class as static variables?

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
-    halDMA.Init.MemInc              = DMA_MINC_ENABLE;
-    halDMA.Init.PeriphDataAlignment = DMA_PDATAALIGN_HALFWORD;
-    halDMA.Init.MemDataAlignment    = DMA_MDATAALIGN_HALFWORD;
-    halDMA.Init.Mode                = DMA_CIRCULAR;
-    halDMA.Init.Priority            = DMA_PRIORITY_VERY_HIGH;
-    halDMA.Init.FIFOMode            = DMA_FIFOMODE_DISABLE;
-
-    HAL_DMA_Init(&halDMA);
-    __HAL_LINKDMA(&halADC, DMA_Handle, halDMA);
+    uint8_t adcNum                 = getADCNum();
```
mjmagee991, line 165: This could be a local variable instead of being gotten here

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+    case ADC2_SLOT:
+        dma->Instance     = DMA2_Stream2;
+        dma->Init.Channel = DMA_CHANNEL_1;
+        break;
+    case ADC3_SLOT:
+        dma->Instance     = DMA2_Stream1;
+        dma->Init.Channel = DMA_CHANNEL_2;
+        break;
+    default:
+        return; // Should never get here
```
mjmagee991, line 183: Log an error

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+    }
+    __HAL_LINKDMA(&adcState->halADC, DMA_Handle, *dma);
 }
 
 void ADCf4xx::addChannel(uint8_t rank) {
     GPIO_InitTypeDef gpioInit;
-    Pin myPins[]      = {pin};
     uint8_t numOfPins = 1;
+    uint32_t channel;
+    Pin myPins[]                   = {pin};
```
mjmagee991, line 222: Get rid of `myPins`. That's a terrible variable name

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
+#define ADC2SHIFT 6
+#define ADC3SHIFT 7
 
-// Init static member variables
-ADC_HandleTypeDef ADCf4xx::halADC = {0};
-Pin ADCf4xx::channels[MAX_CHANNELS];
-uint16_t ADCf4xx::buffer[MAX_CHANNELS];
-DMA_HandleTypeDef ADCf4xx::halDMA = {0};
+// Combines the channel memory value with the ADC peripherals it supports into one uint32_t
+#define CHANNEL_SET(adc1, adc2, adc3, ch) (ch | (adc1 << ADC1SHIFT) | (adc2 << ADC2SHIFT) | (adc3 << ADC3SHIFT))
```
mjmagee991, line 59: Replace this with a `constexpr` function. It's like a macro in that it runs at compile time, but it's strongly typed, so it's easier to understand

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+    case ADCPeriph::ONE:
+        return ADC1_SLOT;
+    case ADCPeriph::TWO:
+        return ADC2_SLOT;
+    case ADCPeriph::THREE:
+        return ADC3_SLOT;
+    }
+}
+
+void ADCf4xx::InitTimer() {
```
mjmagee991, line 328: Can you explain what exactly the timer is doing here?

mjmagee991 (changes requested):
A few questions

mjmagee991:
@mjh9585 I don't think it needs to be addressed now, but I feel like there should be a class that manages the resources for a single internal ADC peripheral instead of a bunch of static arrays

include/core/io/platform/f4xx/ADCf4xx.hpp
```diff
@@ -25,22 +31,72 @@ class ADCf4xx : public ADC {
...
-    /// The ADC will be initialized once then each channel will be added on.
-    static ADC_HandleTypeDef halADC;
-    /// Static list of all channels supported by the ADC
-    static Pin channels[MAX_CHANNELS];
-    /// Buffer for DMA where each spot represents the value read in from a
-    /// channel
-    static uint16_t buffer[MAX_CHANNELS];
-    static DMA_HandleTypeDef halDMA;
+    // Flag to indicate if the timer has been initialized
+    static bool timerInit;
```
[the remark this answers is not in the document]
tmb5932: Ok so yes, only 1 timer for all ADC's, and I am gonna say yes it has to be timer 8. 
In theory no, in practice yes... It should be possible with any timer, but I can only get it to work with timer 3 or 8 which have a Trigger Event "Update Event". Timer 3 is a basic timer and usable by the programmer, but Timer 8 is advanced and not used anywhere else, so it was better than 3, which could be used elsewhere.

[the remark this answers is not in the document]
mjmagee991: So you tried other timers, and they didn't work? Do you know why?

include/core/io/platform/f4xx/ADCf4xx.hpp
```diff
@@ -25,22 +31,72 @@ class ADCf4xx : public ADC {
...
+     * halADC: HAL handle for configuring and controlling the ADC peripheral.
+     * rank: The ADC channel rank in the sequence (starts at 1).
+     * isADCInit: Flag to indicate whether the ADC has been initialized.
+     * channels: Array of pins mapped to ADC channels. Initialized to all Pin::DUMMY.
+     *             The array size is defined by MAX_CHANNELS.
+     * buffer: Array to store ADC values for each channel, indexed according to the channels array.
+     *           Each entry corresponds to a specific ADC channel.
+     * halDMA: HAL handle for configuring and controlling DMA for the ADC.
+     */
+    typedef struct ADC_State {
```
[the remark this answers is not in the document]
tmb5932: Matt and I, respectably, disagree with you. I am using typedef struct so i can give it a type after, so I dont have to do "struct ADC_State" everywhere.

include/core/manager.hpp
```diff
@@ -145,14 +145,14 @@ namespace core::io {
  * @param[in] pin The pin to use with the ADC
  */
 #ifdef ADC_SUPPORTED
-template<Pin pin>
+template<Pin pin, ADCPeriph adcPeriph = ADCPeriph::ONE>
```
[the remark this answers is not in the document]
tmb5932: Yes, one is the default and "master" adc, at least for the f3's and f4, so if your ever going to just use 1 adc, you want adc1.

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: ahahaha I never noticed thats what it was called. Thats grandfathered in from f3 adc

include/core/io/platform/f4xx/ADCf4xx.hpp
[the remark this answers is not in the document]
tmb5932: So I couldn't get them to work, as I couldn't get the ADC to trigger on Timer Trigger Event's besides Update Event. 
My research says that it is possible, but doesn't give a way to do it, and my testing and attempts didn't get them to work.

include/core/io/platform/f4xx/ADCf4xx.hpp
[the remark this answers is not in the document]
tmb5932: ok so i was corrected by ethan, that I can not use typedef, and then not say struct and it will be inferred. thats my bad :)

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
+    HAL_ADC_IRQHandler(adcHandle[1]);
+}
+
+/**
+ * @brief This function handles DMA2 stream1 global interrupt. (For ADC 3)
+ */
+extern "C" void DMA2_Stream1_IRQHandler(void) {
+    HAL_DMA_IRQHandler(dmaHandle[2]);
+    HAL_ADC_IRQHandler(adcHandle[2]);
+}
```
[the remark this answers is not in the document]
tmb5932: Because STM is mean and confusing. I put them in this order when I wrote it because I went in adc order, not stream order.
Stream 0 handles ADC1.
Stream 1 handles ADC3.
Stream 2 handles ADC2.

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
     }
 
+    // This checks if the pin being used supports the ADC being used
+    if (checkSupport(adcPeriph, channel)) {
+        // Masks channel back to proper value (Zero's out ADC information bits)
+        adcChannel.Channel = channel & 0x1F;
+    } else {
+        log::LOGGER.log(log::Logger::LogLevel::ERROR, "ADC %d DOES NOT SUPPORT PIN 0x%x!!", (adcNum + 1), pin);
+        // Causes HARD FAULT if pin does not support the ADC peripheral being used. THIS IS INTENTIONAL!
+        *((volatile int*) 0xFFFFFFFF) = 0; // This address is invalid
```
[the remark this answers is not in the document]
tmb5932: Ok so the problem is that if I dont hard fault, the programmer (without logger enabled) won't be able to tell why the adc is just reading 0's, as it wont break or anything, it will keep going as if everything worked. 
So should i just like, log it and let it run? Whats ur suggestion

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
-uint16_t ADCf4xx::buffer[MAX_CHANNELS];
-DMA_HandleTypeDef ADCf4xx::halDMA = {0};
+// Combines the channel memory value with the ADC peripherals it supports into one uint32_t
+#define CHANNEL_SET(adc1, adc2, adc3, ch) (ch | (adc1 << ADC1SHIFT) | (adc2 << ADC2SHIFT) | (adc3 << ADC3SHIFT))
 
-ADCf4xx::ADCf4xx(Pin pin) : ADC(pin) {
-    // Flag representing if the ADC has been configured yet
-    static bool halADCisInit = false;
+ADCf4xx::ADC_State_t ADCf4xx::adcArray[3];
+bool ADCf4xx::timerInit = false;
```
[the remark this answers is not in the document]
tmb5932: so I can't set it to be false in the header because it says "Non-const static data member must be initialized out of line", so i did it there... where should it go

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
 
-    // "Rank" represents the order in which the channels are added
-    // Also represents the total number of added channels
-    static uint8_t rank = 1;
+ADCf4xx::ADCf4xx(Pin pin, ADCPeriph adcPeriph) : ADC(pin, adcPeriph) {
+    // Get adc state being updated
+    uint8_t adcNum                 = getADCNum();
+    ADCf4xx::ADC_State_t* adcState = &adcArray[adcNum];
```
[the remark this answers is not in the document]
tmb5932: ok so i removed the adcf4xx:: but likeeee... how much better is ADC_State_t&, cause if i do that I gotta change all the -> to . :|

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
+    }
+
+    if (adcState->isADCInit) {
+        HAL_ADC_Stop_DMA(&adcState->halADC);
+        HAL_DMA_DeInit(&adcState->halDMA);
     } else {
         __HAL_RCC_DMA2_CLK_ENABLE();
-        halADCisInit = true;
+        adcState->isADCInit = true;
     }
```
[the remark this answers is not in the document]
tmb5932: I'm kinda confused by this comment. The isADCInit is unique to each adc instance, so it will be false for the first time that the specific adc peripheral is started (1, 2, or 3), and then any channels added to that specific adc after that it will do the true part of the if statement, which stops the ADC & associated DMA, and then it restarts the ADC and DMA in the constructor later.

If your talking about the __DMA2_CLK_ENABLE(), that just sets a bit to 1, so you can call it as many times as we want, and it will be fine...

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,105 @@
...
 
-    addChannel(rank);
-
-    initADC(rank);
-
     initDMA();
+    initADC(adcState->rank);
+    addChannel(adcState->rank);
```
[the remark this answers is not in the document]
tmb5932: No, it works. I believe at one point actually it wasn't working BECAUSE it was in the channel then adc then dma order

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +127,222 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+        __HAL_RCC_ADC2_CLK_ENABLE();
+        break;
+    case ADC3_SLOT:
+        halADC->Instance = ADC3;
+        __HAL_RCC_ADC3_CLK_ENABLE();
+        break;
+    default:
+        return; // Should never get here
+    }
+    HAL_ADC_Init(halADC);
```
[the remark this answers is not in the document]
tmb5932: every time, because every time we add a channel / instance, we need to initialize the ADC peripheral with a new  number of conversions (1 higher). The number of conversions is the the num of channels on that specific adc.

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: your so smart fr

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: like, are you personally confused about it, or do you literally mean add a comment. im gonna assume its add a comment, but i can answer if u need

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
tmb5932: they need to be accessible by the interrupt handlers which cannot reference the class

src/core/io/platform/f4xx/ADCf4xx.cpp
mjmagee991: I am personally confused, but a comment would probably help too. Put the comment in the header file though

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
mjmagee991: Add a comment explaining that it's dumb

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
mjmagee991: I think you should log at error level and let it go. If you turn the logger off, that's your problem. @mjh9585 what do you think?

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: The timer limits the rate a which the ADC does a conversion, as without the timer slowing it down, it will do it so fast (even on slowest sample speed) that DMA interrupts will be called constantly (this is why the DMA ADC wouldn't work originally on f4). 
So this method initializes the timer that sends a Trigger Event to the ADC's to do a conversion. All ADC's are on the same timer and read at the same time.

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
mjmagee991: I guess this is fine then. I looked it up, and the rules make sense to me

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
mjmagee991: Making it a reference means nobody can ever accidentally make it a null pointer, which is worth doing

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
mjmagee991: For some reason, I thought `isADCInit` was getting set to `false` somewhere. I think this should be fine

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
tmb5932: ok 😞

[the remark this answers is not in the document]
tmb5932: Yo magee if I make it a reference, and also put it into the header so that I dont need to do "ADC_State_t& adcState = adcArray[adcNum];" in every method, then I need to have it passed in as a parameter the constructor. 

But for it to be a parameter in the constructor then I need to have the parameter in the constructor in the hpp file, but the constructor in the public section above the declaration of the struct, so it doesn't see the struct existing. Honestly idk what im doing w these in the hpp, I don't do stuff like this. Should I just like, make a second private section at the top of the header so its above the constructor, or what...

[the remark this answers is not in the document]
mjmagee991: Why would you pass it as a parameter to the constructor? Don't you figure out which one it should be based on  the result of `getADCNum()`?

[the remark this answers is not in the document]
tmb5932: yes. So because of one of ur comments elsewhere, I moved the `ADC_State_t* adcState;` to the hpp so that I don't need to call and set it in every function. 
But if I try to change it to a reference (`ADC_State_t& adcState;`) in the hpp, it give the error 
"Constructor for 'core::io::ADCf4xx' must explicitly initialize the reference member 'adcState'" and tells me to add adcState to the parameters for the constructor. I'm not sure if there is another way to do it but that is what it tells me in the error

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: In the initADC() method on line 139, the ADC's external conversion trigger is set to Timer 8's trigger output. That, along with line 136 and 143 that sets the ADC to only do 1 conversion each trigger, means that it does exactly 1 conversion each timer cycle.

The ADC's listen for the timer trigger outputs. The timer has no awareness of the adc's. It just sends out a Trigger Output every cycle, unaware that anything is using it.

src/core/io/platform/f4xx/ADCf4xx.cpp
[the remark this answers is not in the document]
tmb5932: oh wait, I think I got it to work. i didnt know you could just have whatever values you wanted in the member initialization list. I thought they had to be variables from the parameters... I added it and will push, and have u look at it again to make sure this is right

src/core/io/platform/f4xx/ADCf4xx.cpp
mjmagee991: You should add a nice big block comment around there explaining exactly that. My eyes usually glaze over when looking at the configuration macros, so I imagine a lot of people are going to miss that

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -1,64 +1,104 @@
...
+constexpr uint8_t ADC2SHIFT = 6;
+constexpr uint8_t ADC3SHIFT = 7;
 
-ADCf4xx::ADCf4xx(Pin pin) : ADC(pin) {
-    // Flag representing if the ADC has been configured yet
-    static bool halADCisInit = false;
+// Combines the channel memory value with the ADC peripherals it supports into one uint32_t
+constexpr uint32_t CHANNEL_SET(uint8_t adc1, uint8_t adc2, uint8_t adc3, uint32_t ch) {
+    return (ch | (adc1 << ADC1SHIFT) | (adc2 << ADC2SHIFT) | (adc3 << ADC3SHIFT));
+}
```
mjmagee991, line 62: @mjh9585 How do we like replacing macros with `constexpr` functions? It's something I learned on co-op, and I really like using them

mjmagee991: I talked to Helldog for a while, and we came to the conclusion that `constexpr` functions can potentially problematic, but they're something we should look into in the future. For now, we agreed that it would be better to get rid of this entirely and use a packed bit field struct. This will have the memory savings of packing the bits ourselves, and it also allows us to get rid of the ugly bit masking stuff

include/core/io/platform/f4xx/ADCf4xx.hpp
```diff
@@ -25,22 +32,79 @@ class ADCf4xx : public ADC {
...
+     * halADC: HAL handle for configuring and controlling the ADC peripheral.
+     * rank: The ADC channel rank in the sequence (starts at 1).
+     * isADCInit: Flag to indicate whether the ADC has been initialized.
+     * channels: Array of pins mapped to ADC channels. Initialized to all Pin::DUMMY.
+     *             The array size is defined by MAX_CHANNELS.
+     * buffer: Array to store ADC values for each channel, indexed according to the channels array.
+     *           Each entry corresponds to a specific ADC channel.
+     * halDMA: HAL handle for configuring and controlling DMA for the ADC.
+     */
+    struct ADC_State_t {
```
mjmagee991, line 62: We should follow our normal naming convention
```suggestion
    struct ADCState {
```

include/core/io/platform/f4xx/ADCf4xx.hpp
```diff
@@ -25,22 +32,79 @@ class ADCf4xx : public ADC {
...
+
+    // Array of all ADC peripheral states
+    static ADC_State_t adcArray[NUM_ADCS];
+    // The ADC peripheral state of the current object
+    ADC_State_t& adcState;
+    // The number ADC peripheral which is being used in the current object
+    uint8_t adcNum;
+
+    /**
+     * Checks if the channel that is being initialized supports the ADC peripheral that it is being initialized on.
```
mjmagee991, line 94: Function summaries should be written as commands
```suggestion
     * Check if the channel that is being initialized supports the ADC peripheral that it is being initialized on.
```

samples/adc/multi_adc.cpp
```diff
@@ -17,23 +18,31 @@ int main() {
...
-        uart.printf("--------------------\r\n");
-        uart.printf("ADC0 : %d mV\r\n", static_cast<uint32_t>(adc0.read() * 1000));
-        uart.printf("ADC0: %d%%\r\n", static_cast<uint32_t>(adc0.readPercentage() * 100));
-        uart.printf("ADC0 raw: %d\r\n\r\n", adc0.readRaw());
-
-        uart.printf("ADC1 : %d mV\r\n", static_cast<uint32_t>(adc1.read() * 1000));
-        uart.printf("ADC1: %d%%\r\n", static_cast<uint32_t>(adc1.readPercentage() * 100));
-        uart.printf("ADC1 raw: %d\r\n", adc1.readRaw());
-        uart.printf("--------------------\r\n\r\n");
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "--------------------\r\n");
```
mjmagee991, line 33: The logger adds one `\r\n`, so you should remove one from the end of each of these prints

samples/adc/single_adc.cpp
```diff
@@ -17,18 +18,24 @@ int main() {
...
-    io::ADC& adc0 = io::getADC<io::Pin::PA_0>();
+    io::ADC& adc0 = io::getADC<io::Pin::PA_0, io::ADCPeriph::ONE>();
 
     while (1) {
-        uart.printf("--------------------\r\n");
-        uart.printf("ADC0 : %d mV\r\n", static_cast<uint32_t>(adc0.read() * 1000));
-        uart.printf("ADC0: %d%%\r\n", static_cast<uint32_t>(adc0.readPercentage() * 100));
-        uart.printf("ADC0 raw: %d\r\n\r\n", adc0.readRaw());
-        uart.printf("--------------------\r\n\r\n");
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "--------------------\r\n");
```
mjmagee991, line 32: Same deal with `\r\n`

src/core/io/platform/f3xx/ADCf3xx.cpp
```diff
@@ -22,14 +22,20 @@ extern "C" void DMA1_Channel1_IRQHandler(void) {
 }
 
 namespace core::io {
```
mjmagee991, line 24:
```suggestion
namespace core::io {

```

src/core/io/platform/f4xx/ADCf4xx.cpp
mjmagee991: Make sure you also update F3 to be the same

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +124,225 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+    case ADCPeriph::ONE:
+        return channel & (1 << ADC1SHIFT);
+    case ADCPeriph::TWO:
+        return channel & (1 << ADC2SHIFT);
+    case ADCPeriph::THREE:
+        return channel & (1 << ADC3SHIFT);
+    }
+}
+
+inline uint8_t ADCf4xx::getADCNum(ADCPeriph periph) {
```
mjmagee991, line 312: Why do you think these functions should be inline? Consider what's said here: https://stackoverflow.com/questions/1932311/when-to-use-the-inline-function-and-when-not-to-use-it

mjmagee991 (approved):
A couple last things, but looks pretty good

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: Ok so I originally was thinking that the getADCNum should be inline because it effectively just translates the ADCPeriph enum into an uint8_t, and inlining is good for small and often called function, but now that it is only called twice, both in the constructor, it doesn't really matter. 
For the checkSupport, that was originally inline because of it being just a switch statement, but looking at that stack overflow discussion I think it shouldn't be, just because that is one of the areas we are going to want to have a breakpoint when something doesn't work, and it is only called once per adc channel initialization. Also because the stack overflow mentions avoiding inline when working with binary, which checkSupport() does, so thats another reason.

In all, I'll remove inline from checkSupport(), but im inbetween for getADCNum(), so if you prefer avoiding inline, I can remove it aswell.

src/core/io/platform/f4xx/ADCf4xx.cpp
tmb5932: I'll be honest, ion know how packed bit structs are done in c++. I know its just like using all the bits in a struct so that there isn't empty/wasted bits inbetween variables. I also know there is an impact on the order of the variables you put into the struct but thats all i got. I'll google n look into packed bit structs and how to make/define them in c++. 

Just to clarify, I am creating a struct for the channel uint and the adc support bits, and then have that struct be created instead of calling this CHANNEL_SET() constexpr? 
What about the ADC1_SLOT and the other slots, which are also constexprs. Should I change them back into define macros, so we arent using any constexpr?

tmb5932: The other question I have is, with the packed bit struct, should I have the channel be only 5 bits, because the largest the channel ever is is only 5 bits, or should I give it the full value that it starts as (uint32_t)

src/core/io/platform/f4xx/ADCf4xx.cpp
mjmagee991: I'd prefer to get rid of it unless we have a really good reason to keep it, which we don't in this case

src/core/io/platform/f4xx/ADCf4xx.cpp
mjmagee991: The `constexpr` variables are good as they are. The concern is the `constexpr` macro.

You should be able to put the packed struct into a single byte, and that would be ideal in terms of memory, so you should try to do that

include/core/io/pin.hpp
Ethan-Caracoglia, line 1: Good change

tmb5932:
Just tested f4 and f3 adc single and multi (tested with diff combos all the way up to all 3 adc periphs with 3 channels on each), and everything works. Helldog saw it 👨‍🍳

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +117,224 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+/*
+ * This method initializes timer 8 with a Trigger Output of "Update Trigger", with a timer frequency of 1kHz.
+ * The timer does not specifically tell the exact ADC to do a conversion, it just sends a general Update Trigger every
+ * cycle. The ADC's are listening for this Update Trigger from timer 8, which is set in the ADC initialization.
+ */
+void ADCf4xx::initTimer() {
+    TIM_ClockConfigTypeDef sClockSourceConfig = {0};
+    TIM_MasterConfigTypeDef sMasterConfig     = {0};
+
+    ADCf4xx::htim8.Instance               = TIM8;
```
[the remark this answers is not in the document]
tmb5932: because the timer is static and without the ADCf4xx:: in front of it it wont build cause it tries to call it as the instances htim, but that doesnt exist. The other places where I added the ADCf4xx are for the same reason.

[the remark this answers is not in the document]
mjmagee991: Ok, sounds good

samples/adc/multi_adc.cpp
```diff
@@ -17,23 +18,30 @@ int main() {
...
+    core::log::LOGGER.setLogLevel(core::log::Logger::LogLevel::INFO);
+
     uart.printf("Starting ADC test\r\n");
 
     time::wait(500);
 
-    io::ADC& adc0 = io::getADC<io::Pin::PA_0>();
-    io::ADC& adc1 = io::getADC<io::Pin::PA_1>();
+    io::ADC& adc0 = io::getADC<io::Pin::PA_0, io::ADCPeriph::ONE>();
+    io::ADC& adc1 = io::getADC<io::Pin::PC_4, io::ADCPeriph::ONE>();
```
mjh9585, line 30: why is only one ADC used in the multi-ADC sample?

samples/adc/multi_adc.cpp
```diff
@@ -17,23 +18,30 @@ int main() {
...
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC1 : %d mV", static_cast<uint32_t>(adc0.read() * 1000));
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC1: %d%%", static_cast<uint32_t>(adc0.readPercentage() * 100));
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "ADC1 raw: %d\r\n", adc0.readRaw());
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC1 : %d mV", static_cast<uint32_t>(adc1.read() * 1000));
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC1: %d%%", static_cast<uint32_t>(adc1.readPercentage() * 100));
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "ADC1 raw: %d\r\n", adc1.readRaw());
```
mjh9585, line 43: Should this extra new line be here? Same for the ADC above. Also both of your ADCs print with the same identifier, how are you suppose to know which is which?
```suggestion
        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "ADC1 raw: %d", adc1.readRaw());
```

samples/adc/single_adc.cpp
```diff
@@ -17,18 +18,24 @@ int main() {
...
-        uart.printf("ADC0 : %d mV\r\n", static_cast<uint32_t>(adc0.read() * 1000));
-        uart.printf("ADC0: %d%%\r\n", static_cast<uint32_t>(adc0.readPercentage() * 100));
-        uart.printf("ADC0 raw: %d\r\n\r\n", adc0.readRaw());
-        uart.printf("--------------------\r\n\r\n");
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "--------------------");
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC0 : %d mV", static_cast<uint32_t>(adc0.read() * 1000));
+        core::log::LOGGER.log(
+            core::log::Logger::LogLevel::INFO, "ADC0: %d%%", static_cast<uint32_t>(adc0.readPercentage() * 100));
+        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "ADC0 raw: %d\r\n", adc0.readRaw());
```
mjh9585, line 37: Should this extra new line be here?
```suggestion
        core::log::LOGGER.log(core::log::Logger::LogLevel::INFO, "ADC0 raw: %d", adc0.readRaw());
```

src/core/io/platform/f3xx/ADCf3xx.cpp
```diff
@@ -38,6 +38,7 @@ ADCf3xx::ADCf3xx(Pin pin) : ADC(pin) {
 
     // Maximum number of ADC channels have already been added
     if (rank == MAX_CHANNELS) {
+        log::LOGGER.log(log::Logger::LogLevel::WARNING, "ADC1 ALREADY HAS MAX NUMBER OF CHANNELS!!");
```
mjh9585, line 41: Should be an error since the class is not fully initialized.

src/core/io/platform/f3xx/ADCf3xx.cpp
```diff
@@ -137,60 +136,70 @@ void ADCf3xx::addChannel(uint8_t rank) {
...
     }
 
     // Subtract 1 because rank starts at 1
     channels[rank - 1] = pin;
 
+    // This checks if the pin being used supports the ADC being used
+    if (checkSupport(adcPeriph, channelStruct)) {
+        adcChannel.Channel = channelStruct.channel;
+    } else {
+        log::LOGGER.log(log::Logger::LogLevel::ERROR, "DOES NOT SUPPORT PIN 0x%x!!", pin);
```
mjh9585, line 200: This should also probably stop since we are in a bad state.

src/core/io/platform/f3xx/ADCf3xx.cpp
```diff
@@ -7,10 +7,10 @@
 #include <HALf3/stm32f3xx_hal_adc.h>
 #include <HALf3/stm32f3xx_hal_adc_ex.h>
 #include <core/platform/f3xx/stm32f3xx.hpp>
+#include <core/utils/log.hpp>
 
 namespace {
-/// This is made as a global variable so that it is accessible in the
-// interrupt.
+/// This is made as a global variable so that it is accessible in the interrupt.
```
mjh9585, line 13: This anonymous namespace should be in the `core::io` namespace and include the ISR

src/core/io/platform/f4xx/ADCf4xx.cpp
```diff
@@ -84,109 +117,224 @@ void ADCf4xx::initADC(uint8_t num_channels) {
...
+        channelStruct = {}; // sets all values to 0
+        log::LOGGER.log(log::Logger::LogLevel::ERROR, "INVALID PIN 0x%x!!", pin);
         break; // Should never get here
     }
 
+    // This checks if the pin being used supports the ADC being used
+    if (checkSupport(adcPeriph, channelStruct)) {
+        adcChannel.Channel = channelStruct.channel;
+    } else {
+        log::LOGGER.log(log::Logger::LogLevel::ERROR, "ADC %d DOES NOT SUPPORT PIN 0x%x!!", (adcNum + 1), pin);
```
mjh9585, line 279: This should probably stop if we have a bad pin.

mjh9585 (approved):
Looks pretty good, just a few little things.

src/core/io/platform/f3xx/ADCf3xx.cpp
tmb5932: What do you mean by the class is not fully initialized? the rank is only max if there are already 15 channels open on the ADC.

src/core/io/platform/f3xx/ADCf3xx.cpp
tmb5932: stop how? if your saying you want it to crash, me and magee talked about it in another comment and decided against hard faulting.

src/core/io/platform/f3xx/ADCf3xx.cpp
tmb5932: i thought the whole point was that the namespace was outside of the core::io? your saying it should be 
`namespace core::io {`
`namespace {`
`extern "C" void DMA1_Channel1_IRQHandler() {}`
`}`
`... other code here ...`
`}`

src/core/io/platform/f3xx/ADCf3xx.cpp
mjmagee991: I think he means that this instance of the class won't be fully initialized, so it's not going to behave properly. Therefore, you should report an error instead of a warning

PR #110, opened 2024-10-24
FWT-73 Variety is the Driver Fixes of Life

aclowmclaughlin:
Went to fix F4 UART missing switch statement and got carried away. Most compilation warnings are gone, including a missing SPI method, obviously the missing uart switch statement, a random encoder driver method fix (wow wonder who made that method originally).

### DIFF
--- a/src/core/dev/Encoder.cpp
+++ b/src/core/dev/Encoder.cpp
@@ -98,7 +98,7 @@ void Encoder::setRangeAndPosition(uint32_t newRange, uint32_t newPosition) {
 }
 
 int8_t Encoder::readPinValues() {
-    //we are storing these as uints so we can combine them easier.
+    // we are storing these as uints so we can combine them easier.
     uint8_t aPos = a.readPin() == io::GPIO::State::HIGH;
     uint8_t bPos = b.readPin() == io::GPIO::State::HIGH;
     // Calculating and returning the position.
@@ -114,7 +114,7 @@ int8_t Encoder::readPinValues() {
     case 0b10:
         return 3;
     }
-    //This will never happen, the switch statement covers all possible values of comb
+    // This will never happen, the switch statement covers all possible values of comb
     return 0;
 }
 
--- a/src/core/dev/Encoder.cpp
+++ b/src/core/dev/Encoder.cpp
@@ -28,7 +28,7 @@ void Encoder::aInterruptHandler() {
         log::LOGGER.log(log::Logger::LogLevel::DEBUG, "aInterrupt Skipped, reason: on cooldown");
         return;
     }
-    int8_t change;
+    int8_t change = 0;
     switch (currentRelPos) {
     case 0:
         change        = 1;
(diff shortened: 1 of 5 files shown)


src/core/dev/Encoder.cpp
```diff
@@ -114,17 +114,23 @@ void Encoder::setRangeAndPosition(uint32_t newRange, uint32_t newPosition) {
 }
 
 int8_t Encoder::readPinValues() {
-    bool aPos = (bool) a.readPin();
-    bool bPos = (bool) b.readPin();
-    // calculating and returning the position
-    if (aPos == 0 && bPos == 0) {
+    uint8_t aPos = (bool) a.readPin();
```
mjmagee991, line 117: Why did you do this?

src/core/dev/Encoder.cpp
```diff
@@ -114,17 +114,23 @@ void Encoder::setRangeAndPosition(uint32_t newRange, uint32_t newPosition) {
...
         return 1;
-    } else if (aPos == 1 && bPos == 1) {
+    case 0b11:
         return 2;
-    } else if (aPos == 0 && bPos == 1) {
+    case 0b10:
         return 3;
+    default:
+        // this default case will never really happen.
+        return 0;
```
mjmagee991, line 133: If this should never happen, make it return a value that is recognizably wrong, like -1

mjmagee991 (changes requested):
A couple questions

src/core/dev/Encoder.cpp
aclowmclaughlin: The switch statement is more efficient, personally I think more readable, and having the default case removes the compiler warning that the method may not return a value.

src/core/dev/Encoder.cpp
aclowmclaughlin: I considered that, but the issue is if this fails, I think failing in a way that just doesn't register the encoder moving is far safer than failing in a way that causes the encoder to move super far makes more sense. Maybe I should add a log statement for the default case, but the issue is having it return -1 will not actually break anything, so it will still fail silently. Also it's literally impossible for the default case to happen. Like there are only 4 possible values that can even be calculated.

mjh9585 (approved):
Looks good, just one change I would suggest.

src/core/dev/Encoder.cpp
mjmagee991: Can you add a comment here explaining why you're keeping it as a `uint8_t`? Your explanation makes sense, but it's not immediately obvious. Also, you may as well not cast to `bool` here because you're immediately converting to `uint8_t`

src/core/dev/Encoder.cpp
mjmagee991: The only way for it to get to the default case is if someone accidentally deletes one of the other cases, which is highly unlikely. However, I would argue that when someone does something really dumb, we should be loud about it.

For example if someone deletes the case for `0b10` and the default case returns 0, someone might think the encoder is just skipping position 3 for some reason and spend time debugging that. If instead it returned -1, it would be immediately obvious that something major is wrong with the driver

mjmagee991 (changes requested):

src/core/dev/Encoder.cpp
aclowmclaughlin: Cleaned it up and made it a bit more obvious what is happening.

src/core/dev/Encoder.cpp
aclowmclaughlin: Removed the default case and added a return 0; line outside of the switch statement to prevent the compiler complaining. Once again this return 0; outside of the switch will never happen.

PR #111, opened 2024-11-04
FWT-16 We CAN do 2 For 4

mjh9585:
Adds dual CAN bus support for STM32F446xx in EVT-core.

- Adds base CAN bus support to STM32F446xx 
    - Support CAN 1 hardware
    - Support CAN 2 hardware
- Allows simultaneous CAN bus usage on both interfaces

Screenshot 2024-11-03 224051

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Build files
-build/
+build*/
 html/
 latex/
 compile_commands.json
--- a/include/core/io/CAN.hpp
+++ b/include/core/io/CAN.hpp
@@ -133,12 +133,12 @@ protected:
     Pin txPin;
     /** The CAN receive pin */
     Pin rxPin;
-    /** Function pointer to call for the interrupt handler */
-    void (*handler)(CANMessage&, void* priv);
-    /** Private data to pass into the IRQ handler */
-    void* priv;
     /** If CAN should operate in loop back mode */
     bool loopbackEnabled;
+    /** Function pointer to call for the interrupt handler */
+    void (*handler)(CANMessage&, void* priv);
+    /** Context to pass into the IRQ handler */
+    void* priv;
 };
 
 } // namespace core::io
--- a/include/core/io/CAN.hpp
+++ b/include/core/io/CAN.hpp
@@ -128,15 +128,11 @@ public:
      */
     static constexpr uint32_t DEFAULT_BAUD = 500000;
 
-private:
+protected:
     /** The CAN transmit pin */
     Pin txPin;
     /** The CAN receive pin */
     Pin rxPin;
-    /** Represents if filtering should take place for CAN ids */
-    bool filtering;
-
-protected:
     /** Function pointer to call for the interrupt handler */
     void (*handler)(CANMessage&, void* priv);
     /** Private data to pass into the IRQ handler */
--- a/include/core/io/platform/f4xx/CANf4xx.hpp
+++ b/include/core/io/platform/f4xx/CANf4xx.hpp
@@ -17,8 +17,8 @@
 namespace core::io {
 
 /**
- * STMF4xx implementation of the CAN protocol. The STM32f4xx has an on
- * board CAN controller which adds additional features.
+ * STM32f4xx implementation of the CAN protocol. The STM32f4xx has an
+ * onboard CAN controller which adds additional features.
  *
  * 1. Ability to generate interrupts which allows users to add custom call
  * backs
@@ -92,10 +92,22 @@ public:
      */
     bool triggerIRQ(CANMessage& message);
 
+    static constexpr uint8_t EMERGENCY_FILTER_BANK_INDEX = 1;
+
     /** Second filter bank start index*/
     static constexpr uint8_t SECOND_FILTER_BANK_INDEX = 14;
 
 private:
+    /**
+     * Get the CAN ID that is associated with the given pin
+     * combination. This information is pulled from the STM32F446xx
+     * documentation with easier documentation on in CUBEMX
+     *
+     * @param sclPin The selected I2C clock pin
+     * @return The port ID associated with the selected pins
+     */
+    static uint8_t getPortID(Pin txPin, Pin rxPin);
+
     /** Instance of the HAL can interface */
     CAN_HandleTypeDef halCAN;
     /** Queue which holds received CAN messages */
--- a/include/core/io/platform/f4xx/CANf4xx.hpp
+++ b/include/core/io/platform/f4xx/CANf4xx.hpp
@@ -1,0 +1,104 @@
+#ifndef _EVT_CANf4xx_
+#define _EVT_CANf4xx_
+
+#include <stdint.h>
+
+#include <HALf4/stm32f4xx.h>
+#include <HALf4/stm32f4xx_hal_can.h>
+
+#include <core/io/CAN.hpp>
+#include <core/utils/types/FixedQueue.hpp>
+
+// Allows for resizable CAN queue if needed
+#ifndef CAN_MESSAGE_QUEUE_SIZE
+    #define CAN_MESSAGE_QUEUE_SIZE 100
+#endif
+
+namespace core::io {
+
+/**
+ * STMF4xx implementation of the CAN protocol. The STM32f4xx has an on
+ * board CAN controller which adds additional features.
+ *
+ * 1. Ability to generate interrupts which allows users to add custom call
+ * backs
+ *
+ * 2. Hardware based message filtering which allows the filtering of messages
+ * to be handled by hardware not software.
+ */
+class CANf4xx : public CAN {
+public:
+    /**
+     * Create a new instance of an STM32f4xx CAN interface
+     *
+     * @param[in] txPin The pin to trasmit CAN messages on
+     * @param[in] rxPin The pin to receive CAN messages on
+     * @param[in] loopbackEnabled Flag for enabling CAN loop back
+     */
+    CANf4xx(Pin txPin, Pin rxPin, bool loopbackEnabled = false);
+
+    /**
+     * Attempt to join the CAN network.
+     *
+     * For the STM32f4xx this involves attempting to startup the CAN
+     * interface. This could cause an error in the case of invalid
+     * parameters.
+     *
+     * @param[in] autoBusOff Indicates the state halCAN.Init.AutoBusOff should be in
+     *
+     * @return CANStatus::OK on success, CANStatus::ERROR otherwise
+     */
+    CANStatus connect(bool autoBusOff = false) override;
+
+    /**
+     * Disconnect from the CAN network.
+     *
+     * For the STM32f4xx this involves disconnect from the CAN interface.
+     *
+     * @return CANStatus::OK if we successfully de-init the CAN interface
+     */
+    CANStatus disconnect();
+
+    CANStatus transmit(CANMessage& message);
+
+    CANStatus receive(CANMessage* message, bool blocking = false);
+
+    CANStatus addCANFilter(uint16_t filterExplicitId, uint16_t filterMask, uint8_t filterBank);
+
+    CANStatus enableEmergencyFilter(uint32_t state);
+
+    void addIRQHandler(void (*handler)(CANMessage&, void* priv), void* priv);
+
+    /**
+     * Add a message to the CAN receive queue.
+     *
+     * NOTE: This is public for use with the STM HAL interrupt handler. This
+     * method should not be used outside of that application.
+     *
+     * @param[in] message The CANmessage to add to the receive queue
+     */
+    void addCANMessage(CANMessage& message);
+
+    /**
+     * Manually trigger the user specified interrupt handler. This is intended
+     * to be used by the STM HAL interrupt handler and generally should
+     * not be used beyound that use case.
+     *
+     * NOTE: This is public for use with the STM HAL interrupt handler. This
+     * method should not be used outside of that application.
+     *
+     * @param[in] message The message to pass to the interrupt handler
+     * @return True if the interrupt handler exists and has handled the message
+     */
+    bool triggerIRQ(CANMessage& message);
+
+private:
+    /** Instance of the HAL can interface */
+    CAN_HandleTypeDef halCAN;
+    /** Queue which holds received CAN messages */
+    core::types::FixedQueue<CAN_MESSAGE_QUEUE_SIZE, CANMessage> messageQueue;
+};
+
+} // namespace core::io
+
+#endif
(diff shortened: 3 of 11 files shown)


Ethan-Caracoglia (approved):
Looks good

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+ */
+struct CANf4Global {
+    // Pointer to the CANf446xx interface
+    core::io::CANf4xx* canInstance = nullptr;
+
+    // Pointer to the CAN handler for HAL,
+    CAN_HandleTypeDef* canHandler = nullptr;
+};
+
+struct CANf4Global globalCAN[2];
```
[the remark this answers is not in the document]
mjmagee991: Also, the fact that we _have_ to use globals every time we use an interrupt really needs to be addressed. This is gross

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+    core::io::CANf4xx* canInstance;
+    if (hcan == globalCAN[0].canHandler) {
+        canInstance = globalCAN[0].canInstance;
+    } else {
+        canInstance = globalCAN[1].canInstance;
+    }
+
+    // Check to see if a user defined IRQ has been provided
+    if (canInstance->triggerIRQ(message))
+        return;
```
mjmagee991, line 70:
```suggestion
    if (canInstance->triggerIRQ(message)) {
        return;
    }
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+#include <core/io/pin.hpp>
+#include <core/io/platform/f4xx/CANf4xx.hpp>
+#include <core/io/platform/f4xx/GPIOf4xx.hpp>
+#include <core/platform/f4xx/stm32f4xx.hpp>
+#include <core/utils/log.hpp>
+#include <core/utils/time.hpp>
+#include <core/utils/types/FixedQueue.hpp>
+
+namespace {
```
mjmagee991, line 16: Is there a reason not to put this in the same namespace as the other stuff? You can wrap the anonymous namespace with `core::io` and it would behave the same

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * Get the CAN ID that is associated with the given pin
+ * combination. This information is pulled from the STM32F446xx
+ * documentation with easier documentation on in CUBEMX
+ *
+ * @param sclPin The selected I2C clock pin
+ * @return The port ID associated with the selected pins
+ */
+static uint8_t getPortID(Pin txPin, Pin rxPin) {
```
mjmagee991, line 87: Why not declare this in the header, so this comment isn't floating here?

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+    halCAN.Init.Mode              = mode;
+    halCAN.Init.SyncJumpWidth     = CAN_SJW_1TQ;
+    halCAN.Init.TimeSeg1          = CAN_BS1_8TQ;
+    halCAN.Init.TimeSeg2          = CAN_BS2_1TQ;
+    halCAN.Init.TimeTriggeredMode = DISABLE;
+    if (autoBusOff) {
+        halCAN.Init.AutoBusOff = ENABLE;
+    } else {
+        halCAN.Init.AutoBusOff = DISABLE;
+    }
```
mjmagee991, line 180:
```suggestion
    halCAN.Init.AutoBusOff = autoBusOff ? ENABLE : DISABLE;
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+    CAN_TxHeaderTypeDef txHeader                      = {0};
+    uint8_t payload[CANMessage::CAN_MAX_PAYLOAD_SIZE] = {0};
+
+    uint32_t mailbox = CAN_TX_MAILBOX0;
+
+    // Set the message ID
+    if (message.isCANExtended())
+        txHeader.ExtId = message.getId();
+    else
+        txHeader.StdId = message.getId();
```
mjmagee991, line 226:
```suggestion
    if (message.isCANExtended()) {
        txHeader.ExtId = message.getId();
    } else {
        txHeader.StdId = message.getId();
    }
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+CAN::CANStatus CANf4xx::receive(CANMessage* message, bool blocking) {
+    bool hasMessage = false;
+
+    // Check to make sure a message is available, if blocking, wait until
+    // a message is available.
+    do {
+        hasMessage = !messageQueue.isEmpty();
+        // If the user does not want to wait for a message, return nullptr
+        if (!blocking && !hasMessage)
+            return CANStatus::TIMEOUT;
```
mjmagee991, line 271:
```suggestion
        if (!blocking && !hasMessage) {
            return CANStatus::TIMEOUT;
        }
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    if (HAL_CAN_ConfigFilter(&halCAN, &emergencyFilter) != HAL_OK) {
+        return CANStatus::ERROR;
+    }
+    return CANStatus::OK;
+}
+
+void CANf4xx::addCANMessage(CANMessage& message) {
+    if (messageQueue.canInsert())
+        messageQueue.append(message);
```
mjmagee991, line 336:
```suggestion
    if (messageQueue.canInsert()) {
        messageQueue.append(message);
    }
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+bool CANf4xx::triggerIRQ(CANMessage& message) {
+    if (handler == nullptr)
+        return false;
```
mjmagee991, line 341:
```suggestion
    if (handler == nullptr) {
        return false;
    }
```

.gitignore
```diff
@@ -1,5 +1,5 @@
 # Build files
-build/
+build*/
```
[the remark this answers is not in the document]
mjh9585: This allows you to have multiple builds which allows you to setup clion to literally just select what chip you are building for instead of having to edit the cmake args every time you want to build for a different chip. Really useful when constantly switching between chips. This is a good thing.

include/core/io/CAN.hpp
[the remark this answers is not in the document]
mjh9585: Unused variable

samples/can/back_and_forth/main.cpp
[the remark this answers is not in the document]
mjh9585: makes it easier to tell that can is actually alive and the processor hasn't died in some way.

samples/can/multiCAN/main.cpp
```diff
@@ -0,0 +1,122 @@
...
+    core::platform::init();
+    io::UART& uart = io::getUART<io::Pin::UART_TX, io::Pin::UART_RX>(9600);
+    log::LOGGER.setUART(&uart);
+    log::LOGGER.setLogLevel(log::Logger::LogLevel::DEBUG);
+
+    // Get CAN instances
+
+    // CAN 1
+    io::CAN& can1 = io::getCAN<io::Pin::PA_12, io::Pin::PA_11>();
+    // io::CAN& can1 = io::getCAN<io::Pin::PB_9, io::Pin::PB_8>();
```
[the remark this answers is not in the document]
mjh9585: yes, all of those are the valid CAN pins. This provides a helpful reference to someone that is trying to setup CAN or needs an alternative configuration to test it on another board.

samples/can/multiCAN/main.cpp
```diff
@@ -0,0 +1,122 @@
...
+    result = can2.connect();
+    if (result != io::CAN::CANStatus::OK) {
+        uart.printf("[CAN2] Failed to connect to the CAN network\r\n");
+        return 1;
+    }
+
+    while (true) {
+        // Update the payload of the CAN messages with the counts.
+        payload1[7] = can1Count;
+        payload2[5] = can2Count;
```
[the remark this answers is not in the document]
mjh9585: To more clearly differentiate, especially when both are printing to the same terminal.

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    HAL_CAN_GetRxMessage(hcan, CAN_RX_FIFO0, &rxHeader, payload);
+
+    // Construct the CANmessage
+    bool isExtended = rxHeader.IDE == CAN_ID_EXT;
+    uint32_t id     = isExtended ? rxHeader.ExtId : rxHeader.StdId;
+    core::io::CANMessage message(id, rxHeader.DLC, payload, isExtended);
+
+    core::io::CANf4xx* canInstance;
+    if (hcan == globalCAN[0].canHandler) {
```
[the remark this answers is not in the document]
mjh9585: That's fine since the pointer is bound to the physical hardware peripheral address and will never change for the chip. Also this is some of the BS the HAL does.

src/core/io/platform/f4xx/CANf4xx.cpp
mjh9585: Just following the organization standard of the other files in EVT-core. Might be better to put it in the `core::io` to avoid global name conflics.

src/core/io/platform/f4xx/CANf4xx.cpp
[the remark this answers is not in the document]
mjh9585: I don't believe there are any alternatives since interrupts don't have any context and we need context to callback to our class.

include/core/io/CAN.hpp
```diff
@@ -128,15 +128,11 @@ class CAN {
      */
     static constexpr uint32_t DEFAULT_BAUD = 500000;
 
-private:
+protected:
```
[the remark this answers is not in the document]
mjh9585: The port id is only used during initialization so it is a bit of a waste to hold onto it for the duration of the program especially since the pins that are a required part of the abstract class are used to derive that data. Additionally we are storing the pins in the abstract class and never using it, defeating the point of storing them in the first point. In general, I would say an abstract class members should be protected so that a derived class can properly extend the parent's functionality unless that member is critical to the parent's common behavior.

.gitignore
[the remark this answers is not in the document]
mjmagee991: How would you like to write some documentation explaining how to do this?

include/core/io/CAN.hpp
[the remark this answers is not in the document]
mjmagee991: This argument makes me think we should delete the `txPin` and `rxPin` from this class and just save port number in the concrete classes, but that may be more work than it's worth right now

include/core/io/platform/f4xx/CANf4xx.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+    CANStatus transmit(CANMessage& message);
+
+    CANStatus receive(CANMessage* message, bool blocking = false);
+
+    CANStatus addCANFilter(uint16_t filterExplicitId, uint16_t filterMask, uint8_t filterBank);
+
+    CANStatus enableEmergencyFilter(uint32_t state);
+
+    void addIRQHandler(void (*handler)(CANMessage&, void* priv), void* priv);
```
mjmagee991, line 70: At least some of these should be marked override

include/core/io/platform/f4xx/CANf4xx.hpp
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * NOTE: This is public for use with the STM HAL interrupt handler. This
+     * method should not be used outside of that application.
+     *
+     * @param[in] message The message to pass to the interrupt handler
+     * @return True if the interrupt handler exists and has handled the message
+     */
+    bool triggerIRQ(CANMessage& message);
+
+    static constexpr uint8_t EMERGENCY_FILTER_BANK_INDEX = 1;
```
mjmagee991, line 95: Add a Doxygen comment to this

samples/can/multiCAN/README.md
mjmagee991, line 1: Why is this README necessary? Isn't this expressed clearly in the comment at the top of `main.cpp`?

samples/can/multiCAN/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+    uint8_t can1Count = 0;
+    uint8_t can2Count = 0;
+
+    // CAN message that will be sent
+    uint8_t payload1[] = {0xDE, 0xAD, 0xBE, 0xBE, 0xEF, 0x09, 0x08, 0x07};
+    uint8_t payload2[] = {0xBE, 0xEF, 0xBE, 0xDE, 0xAD, 0x00, 0x01, 0x02};
+    io::CANMessage can1TransmitMessage(1, 8, &payload1[0], false);
+    io::CANMessage can2TransmitMessage(2, 8, &payload2[0], false);
+
+    uart.printf("\r\nClock: %d\r\n", SystemCoreClock);
```
mjmagee991, line 83: I would prefer you just use the logger throughout, but that's up to you

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,338 @@
+
+
```
mjmagee991, line 2:
```suggestion
```

src/core/io/platform/f4xx/CANf4xx.cpp
```diff
@@ -0,0 +1,338 @@
...
+    default:
+        log::LOGGER.log(log::Logger::LogLevel::ERROR, "Invalid CAN RX Pin");
+        rxPort = 0;
+        break;
+    }
+#endif
+
+    if (txPort == rxPort) {
+        return txPort;
+    }
```
mjmagee991, line 115: else log an error

mjmagee991 (approved):
Nice work!

include/core/io/CAN.hpp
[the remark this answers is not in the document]
mjh9585: The `txPin` and `rxPin` are more useful than just a port number as they can be used to reconfigure gpio if necessary. Right now that isn't important, but with a different platform it may.

.gitignore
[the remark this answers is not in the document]
mjh9585: Should probably just be a comment on the website, likely under the running code page. Nothing has fundamentally changed, there are just more options now for the build folder.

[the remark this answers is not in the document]
mjh9585: Added as a Jira issue.

PR #113, opened 2025-01-25
Really Tiny One Small Change (to) My Asinine Keywords Excellently 🐛 🔨 

aclowmclaughlin:
The RTOS stuff wasn't building when EVT-core was used as a library in another project, turns out I should have used CMAKE_CURRENT_LIST_DIR instead of CMAKE_SOURCE_DIR. Silly me!
Anyway here's a whole pr for literally less than one line of changes.
Tested one both EVT-core and when using EVT-core as a library, this change makes it build correctly both ways.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ if(USE_RTOS)
     set(THREADX_TOOLCHAIN "gnu")
     add_subdirectory(libs/threadx)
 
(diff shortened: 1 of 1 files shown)


ActuallyTaylor (approved):
Looks good?

mjh9585 (approved):
🤔

PR #114, opened 2025-01-28
Fixing Queue

aclowmclaughlin:
Fixed the include for FixedQueue, changing it from stdint.h to cstddef, which actually works. For some reason in EVT-core it worked the old way even though it should not have, but it doesn't work outside of EVT-core (or at least in VCU, where I encountered this issue).


ishan-renjen (approved):

mjh9585 (approved):
Yay!

PR #117, opened 2025-03-29
FWT-58 Oh no, Its Terminal!

RyanLeifer04:
generic UART terminal to load and execute functions


include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+     * takes a string for the option key, a string for the description text,
+     * a void pointer to the items callback method, and a void pointer to the items context
+     * @param parent pointer to parent node
+     * @param term pointer to terminal instance this item resides within
+     * @param option a string representing the items "key", the char/string used to select it
+     * @param text a short text description/name of an item
+     * @param cb a void pointer to this items callback method
+     * @param ctx a void pointer to any context information for this menu(if none provided, is NULL)
+     */
+    MenuItem(void* parent, void* term, char* option, char* text, callback_t cb, void* ctx);
```
aclowmclaughlin, line 27: Is there a reason these are void* instead of MenuItem pointers? Isn't it always the case that the parent to a MenuItem will be another MenuItem?

aclowmclaughlin, line 27: If many of these are going to be passed a null value commonly, they could be given default values. Default values only work for objects at the end of the initialization list, so perhaps they should be re-ordered so that things that are almost never null are at the front of the list.

include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+     */
+    void* parent;
+
+    /**
+     * terminal this is in
+     */
+    void* term;
+};
+
+class SubMenu : public MenuItem {
```
aclowmclaughlin, line 115: I don't particularly like that these are in the same hpp but have different cpps. Is there any reason for them to be in the same hpp? if there isn't, please move them to separate hpps.

include/core/utils/terminal/terminal.hpp
```diff
@@ -0,0 +1,106 @@
...
+namespace core::utils {
+
+class Terminal {
+public:
+    /**
+     * constructor for terminal object
+     * takes a uart object and sets up the command line terminal
+     * @param uart an instance of a UART object
+     * @param baud 9600 or 115200
+     */
```
aclowmclaughlin, line 19: incorrect doc comment

include/core/utils/terminal/terminal.hpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * enters the given submenu
+     * @param uart the uart instance to use
+     * @param args the arguments provided, used for enter callback, first index is still option string
+     * @param term void pointer to the terminal this is being done inside of(optional to leave as nullptr, just here to
+     * be used with callbacks)
+     */
+    void enterSub(io::UART& uart, char** args, void* term);
```
aclowmclaughlin, line 50: Is there a reason to pass in another uart instance? won't it always be the same one?

include/core/utils/terminal/terminal.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+private:
+    // menu instance
+    Menu* menu;
+
+    // current submenu, if there is one, nullptr otherwise
+    SubMenu* current;
+
+    // true if currently in main menu
+    bool m;
```
aclowmclaughlin, line 96: rename this to inMainMenu or something more explanatory.

src/core/utils/terminal/menu.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void Menu::printStr(core::io::UART& uart) {
+    // loops through item list, stops when it hits an item with an option string of "null"
+    for (int c = 0; c < itemCount; c++) {
+        if (items[c]->getOption() == "null") {
+            return;
+        }
+        items[c]->printStr(uart);
+    }
```
aclowmclaughlin, line 17: Should probably check for the items being null.

src/core/utils/terminal/menu.cpp
```diff
@@ -0,0 +1,72 @@
...
+            break;
+        }
+    }
+    // make sure next space is within bounds
+    if (c >= itemCount) {
+        return;
+    }
+    // if in bounds place in list, if not return nothing
+    items[c - 1] = item;
+}
```
aclowmclaughlin, line 35: I would like for this to inform the user of a failure. Since it has no return type, it should be easy to make it return an enum with like Success/Failure or something.

src/core/utils/terminal/menu.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+void Menu::delItem(int index) {
+    // set desired index to null
+    items[index] = nullptr;
+    // shift every higher index down
+    for (int i = index; i < itemCount - 1; i++) {
+        items[index] = items[index + 1];
+    } // set highest index to null to ensure there is no duplicate final item
+    items[itemCount - 1] = nullptr;
+}
```
aclowmclaughlin, line 52: Have we thought about using a linked list instead? IS there a reason that we need these to be placed within an array?

src/core/utils/terminal/menuItem.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+void MenuItem::printStr(core::io::UART& uart) {
+    // print in form of op | text
+    uart.printf(option);
+    uart.printf("|");
+    uart.printf(text);
+}
+
+void MenuItem::replace(MenuItem* newItem) {
```
aclowmclaughlin, line 17: The issue with this is that from a memory standpoint, this doesn't actually replace one with the other, it just interlinks their fields. I would instead encourage you to either:
1) make a method that simply allows you to pass in options, test, callback, and context.
2) Don't have this be an option, and instead just rely on someone deleting this menuItem and then replacing it with the new one they want, via the menu itself.

src/core/utils/terminal/menuItem.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+bool MenuItem::equals(MenuItem* it2) {
+    // get option string, text, cb, and ctx of 2nd item
+    char* option2  = it2->getOption();
+    char* text2    = it2->getText();
+    callback_t cb2 = it2->getcb();
+    void* ctx2     = it2->getctx();
+    // if any corresponding arent equal, return false
+    if (option != option2 || text != text2 || cb != cb2 || ctx != ctx2) {
```
aclowmclaughlin, line 32: bear in mind that this will literally compare memory addresses, not contents. If that is intended, that's alright, but otherwise this will not work the way you think it does. Additionally, do we necessarily need an equals() method? Is there any point at which we are using it?

include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+     * constructor for sub-menu sub-class
+     * @param parent same as menuitem
+     * @param term same as menuitem
+     * @param option same as menuitem
+     * @param text same as menuitem
+     * @param cb exit behavior callback, leave null for nothing
+     * @param ctx enterence behavior callback void*(leave null for nothing)
+     * @param items list of items in submenu
+     */
+    SubMenu(void* parent, void* term, char* option, char* text, callback_t cb, void* ctx, MenuItem** items);
```
aclowmclaughlin, line 127: I'm struggling to understand what make a subMenu different from a MenuItem. If they aren't significantly different we should consider if we need to care about the distinction between them.

aclowmclaughlin: additionally, if a submenu is also a menu, maybe it should inherit from menu as well?

include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * description/name of item
+     */
+    char* text;
+
+    /**
+     * pointer to callback method for this item
+     */
+    callback_t cb;
```
aclowmclaughlin, line 210: rename this to callback

include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * context for this item, void* because it is of an abstract type
+     */
+    void* ctx;
```
aclowmclaughlin, line 215: rename this to context

include/core/utils/terminal/menuItem.hpp
```diff
@@ -0,0 +1,238 @@
...
+#define EVT_TERM_MENUITEM
+#include <core/io/UART.hpp>
+
+// macro for max initial item count of submenus
+
+// struct used for callback functions, a uart instance to communicate over,
+// a list of input strings(your input from the terminal), and a void*
+// mostly a placeholder to ease handling void* to a function, when you fill with a function make sure the parameters are
+// (UART,char**,void*)
+using callback_t = void (*)(core::io::UART&, char** inputList, void*);
```
aclowmclaughlin, line 11: Big fan of this, we should possibly consider requiring this instead of typedefs in our code, because this is far more readable. However, it should probably be in the namespace.

aclowmclaughlin (changes requested):

include/core/utils/terminal/menu.hpp
```diff
@@ -0,0 +1,79 @@
...
+     * true if every menu item is equal
+     * @param mnu2 the menu to compare to
+     */
+    bool equals(Menu* mnu2);
+
+private:
+    /**
+     * list of all items contained in the menu
+     */
+    MenuItem** items;
```
aclowmclaughlin, line 70: I'm confused about what we're doing here tbh. Since we are capping the size of the list to such a small number, we could just create the array statically, inside this class, instead of having the user pass the array into the object. This will both protect access and also clean up the main code. We may want to look into using templates to set the size of the menu, but I'm not

PR #120, opened 2025-08-25
Very Special Code

mjh9585:
Adds default VScode configurations to support easy development withing VScode. This uses the collection of STM32Cube extensions for VScode to provide building and debugging support without the need for OpenOCD. Supports STM32F302, STM32F334, and STM32F446. 

- Includes a set of recommend extensions to install when repository is cloned
- Adds default launch config
- Adds set of task for building, cleaning, and running code
- Adds a set of quick buttons to the status bar for running tasks
- Adds a set of CMakePresets for all of the different boards

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -42,7 +42,10 @@
 		{
             "type": "shell",
             "label": "Flash project",
-            "command": "${env:STM32_PRG_PATH}\\STM32_Programmer_CLI.exe",
+            "command": "${env:STM32_PRG_PATH}\\STM32_Programmer_CLI",
+			"windows":{
+				"command": "${env:STM32_PRG_PATH}\\STM32_Programmer_CLI.exe"
+			},
             "args": [
                 "--connect",
                 "port=swd",
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,74 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "cmake",
+			"label": "CMake: clean",
+			"command": "clean",
+			"preset": "__defaultBuildPreset__",
+			"problemMatcher": [],
+			"detail": "CMake clean task"
+		},
+		{
+			"type": "cmake",
+			"label": "CMake: build",
+			"command": "build",
+			"targets": [
+				"all"
+			],
+			"preset": "__defaultBuildPreset__",
+			"group": "build",
+			"problemMatcher": [],
+			"detail": "CMake build task"
+		},
+		{
+			"type": "cmake",
+			"label": "CMake: clean rebuild",
+			"command": "cleanRebuild",
+			"targets": [
+				"all"
+			],
+			"preset": "__defaultBuildPreset__",
+			"group": "build",
+			"problemMatcher": [],
+			"detail": "CMake clean rebuild task"
+		},
+		{
+			"type": "shell",
+			"label": "flash debug",
+			"command": "${command:workbench.action.debug.start}",
+			"detail": "Flash code and launch the debugger"
+		},
+		{
+            "type": "shell",
+            "label": "Flash project",
+            "command": "${env:STM32_PRG_PATH}\\STM32_Programmer_CLI.exe",
+            "args": [
+                "--connect",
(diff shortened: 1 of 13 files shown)


include/core/manager.hpp
ActuallyTaylor, line 1: Should this change be in this PR?

samples/timer/main.cpp
ActuallyTaylor, line 1: Same here, this looks like it should be brought in by another PR.

DiegoLHendrix (approved):
Good job

CMakePresets.json
```diff
@@ -0,0 +1,92 @@
...
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "EVT_CORE_LOG_ENABLE": "1",
+        "TARGET_DEV": "STM32F334x8",
+        "FORCE_COLORED_OUTPUT": "1"
+      },
+      "environment": {
+        "GCC_ARM_TOOLS_PATH": "C://EVT/arm-tools/bin/"
```
ActuallyTaylor, line 33: I am not familiar with vscode environment setup so this may not be applicable..

Is there a way to pull this from the user's environment variables? Maybe a fall back to this would be good. The default setting could lead to problems where people think they don't have the tools installed properly, but VSCode is just pointing to the wrong directory.

.vscode/tasks.json
```diff
@@ -0,0 +1,74 @@
...
+		{
+			"type": "shell",
+			"label": "flash debug",
+			"command": "${command:workbench.action.debug.start}",
+			"detail": "Flash code and launch the debugger"
+		},
+		{
+            "type": "shell",
+            "label": "Flash project",
+            "command": "${env:STM32_PRG_PATH}\\STM32_Programmer_CLI.exe",
```
ActuallyTaylor, line 45: Is there a way to do this without an exe? Thinking of Linux and macOS compatibility. Linux seems to be a big area for VSCode users.

ActuallyTaylor (changes requested):
I think this is a great idea :) I have some questions about some included file changes, and a few about the config.

.vscode/tasks.json
ol2764RIT: I tried both approaches, and on Ubuntu 24.04 at least, the STM32 CLI didn’t work nearly as well as OpenOCD + Cortex-Debug. OpenOCD tends to be the better option on Linux since it’s packaged natively on most distros and works out of the box.

CMakePresets.json
mjh9585: Looks like the environment was just not loading right when I was testing it, but now it is pulling it in correctly.

.vscode/tasks.json
mjh9585: Apparently you can add os specific overrides.

include/core/manager.hpp
mjh9585: The build all fails otherwise and was just trying to get it in

include/core/manager.hpp
```diff
@@ -47,6 +47,7 @@
     #define RTC_SUPPORTED
     #define SPI_SUPPORTED
     #define UART_SUPPORTED
+    #define IWDG_SUPPORTED
```
aclowmclaughlin, line 50: Yeah this is a weird change to make in this pr. I fail to see how it relates to anything else in the pr.

samples/timer/main.cpp
```diff
@@ -46,8 +46,14 @@ int main() {
     // Setup the Timer
     dev::Timer& timer2 = dev::getTimer<dev::MCUTimer::Timer2>(500);
     // F4xx does not support Timers 15 & 16, change them to Timer11 & Timer12
+
```
aclowmclaughlin, line 49: This is good, but not the PR for it.

aclowmclaughlin (changes requested):
There's some unrelated changes that I'd prefer be in a different pr, but otherwise I mean yeah sure looks good ig.

aclowmclaughlin (approved):
Fax my guy spit your shit indeed.

PR #124, opened 2025-09-30
VREF-olutionary: VREF support/instantiation for ADCs

ol2764RIT:
Add configurable VREF voltage support to the ADC system for improved measurement accuracy.

The ADC bases its voltage calculations from the VREF pin on the STM32F302R8, but the ADC hardware doesn't implicitly know what the actual voltage at VREF is. Previously, the system assumed a hardcoded 3.3V reference, which could lead to inaccurate voltage readings when using different reference voltages.

This PR adds `setVref(float vref)` and `getVref()` methods to the ADC interface, enabling accurate measurements with various reference voltages (1.8V, 3.3V, 5.0V, etc.). The implementation replaces hardcoded VREF values with configurable runtime settings in both STM32F3xx and F4xx platforms.

```cpp
io::ADC& adc = io::getADC<io::Pin::PA_0, io::ADCPeriph::ONE>();
adc.setVref(5.0f);  // Set 5V reference
float voltage = adc.read();  // Now calculated with correct VREF
```

And do not fret, the default will always be 3.3V so old ADC code will not break

### DIFF
--- a/include/core/io/platform/f3xx/ADCf3xx.hpp
+++ b/include/core/io/platform/f3xx/ADCf3xx.hpp
@@ -27,13 +27,26 @@ public:
 
     float readPercentage();
 
+    void setVref(float vref);
+
+    float getVref() const;
+
+    /**
+     * Perform ADC calibration for improved accuracy
+     * 
+     * @return true if calibration was successful, false otherwise
+     */
+    static bool calibrate();
+
 private:
     // Max number of channels supported by the ADC
     static constexpr uint8_t MAX_CHANNELS = 15;
-    // Positive reference voltage of the ADC.  Needs to be updated based on the hardware configuration
-    static constexpr float VREF_POS = 3.3;
+    // Default positive reference voltage of the ADC.  Can be updated via setVref()
+    static constexpr float DEFAULT_VREF_POS = 3.3;
     // Max value for a 12 bit ADC reading (2^12 - 1)
     static constexpr uint32_t MAX_RAW = 4095;
+    // Current reference voltage (can be modified at runtime)
+    static float vref_voltage;
     /// This is static since the STM32F3xx only has a single ADC which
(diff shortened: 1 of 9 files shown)


include/core/io/platform/f3xx/ADCf3xx.hpp
```diff
@@ -27,13 +27,19 @@ class ADCf3xx : public ADC {
...
+
+    float getVref() const;
+
 private:
     // Max number of channels supported by the ADC
     static constexpr uint8_t MAX_CHANNELS = 15;
-    // Positive reference voltage of the ADC.  Needs to be updated based on the hardware configuration
-    static constexpr float VREF_POS = 3.3;
+    // Default positive reference voltage of the ADC.  Can be updated via setVref()
+    static constexpr float DEFAULT_VREF_POS = 3.3;
```
ActuallyTaylor, line 38: Can this be a `#DEFINE`

samples/adc/vref_demo.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+int main() {
+    // Initialize system
+    core::platform::init();
+
+    io::UART& uart = io::getUART<io::Pin::UART_TX, io::Pin::UART_RX>(9600);
+
+    // Set up the logger to catch errors in ADC creation
+    core::log::LOGGER.setUART(&uart);
+    core::log::LOGGER.setLogLevel(core::log::Logger::LogLevel::INFO);
```
ActuallyTaylor, line 24: Remove the logger here, since the sample uses raw UART

ActuallyTaylor (changes requested):
Looks mostly good! I just see a quick code removal in one of the samples.

EvanHughes-dev (approved):
LGTM

ActuallyTaylor (approved):

PR #126, opened 2025-10-04
Feature/diegolhendrix/bootloader

DiegoLHendrix:
Started creating header and source files for the CAN flashing bootloader.

- Created boot directories and namespace within src and include.
- Fixed spelling mistake in assembly files.
- Added a temporary hal_print sample.
- Removed `#include <iostream>`  from stm32f446xx.cpp file.


include/boot/CAN.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef EVT_BOOTCAN_HPP
+#define EVT_BOOTCAN_HPP
+
+#include <boot/IOInterface.hpp>
+
+class CAN : public IOInterface {
```
ActuallyTaylor, line 6: Some docs explaining how this is fundamentally different from the other `CAN.hpp` file that we have would be nice.

include/boot/Flash.hpp
```diff
@@ -0,0 +1,34 @@
...
+    SUCCESS = 0,
+    FAILURE = 1,
+    FLASH_BUSY = 2,
+    RESOURCE_LOCKED = 3,
+    ACCESS_FAILURE = 4,
+    FAILED_WRITE = 5,
+    FAILED_READ = 6,
+};
+
+class Flash: public IOInterface {
```
ActuallyTaylor, line 18: Is this Flash memory? Or just a class that handles flashing?

include/boot/IOInterface.hpp
```diff
@@ -0,0 +1,34 @@
+#ifndef EVT_IOINTERFACE_HPP
+#define EVT_IOINTERFACE_HPP
+
+#include <cstdint>
+
+class IOInterface {
+public:
+    IOInterface();
+
+    /**
```
ActuallyTaylor, line 10: Great docs :)

include/boot/UART.hpp
```diff
@@ -0,0 +1,16 @@
+#ifndef EVT_BOOTUART_H
+#define EVT_BOOTUART_H
+
+#include <boot/IOInterface.hpp>
+
+class UART: public IOInterface {
```
ActuallyTaylor, line 6: Same as my comment on `CAN.hpp`, this needs docks to explain its difference from the normal `UART.hpp`.

samples/hal_print/main.cpp
ActuallyTaylor, line 1: Can we have a comment in this file explaining what it is? I fear that someone may look at this file and assume it is how we do CAN and UART.

PR #138, opened 2026-04-05
Grabbing Pins In Operation

DannyCato:
(I'm not funny enough to think of anything good)

### DIFF
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -1,8 +1,8 @@
 #ifndef _EVT_GPIO_
 #define _EVT_GPIO_
 
-#include <cstdint>
 #include <core/io/pin.hpp>
+#include <cstdint>
 
 namespace core::io {
 // Forward declarations:
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -19,7 +19,6 @@ namespace core::io {
  */
 class GPIO {
 public:
-#pragma pack(push, 1)
     /**
      * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
      * the value is meant to quickly set or read a value with ease.
@@ -44,7 +43,7 @@ public:
      * then
      * a.pin_12 && a.pin_8 && a.pin_4 && a.pin_0 == 1
      */
-    union PinPack {
+    union __attribute__((packed)) PinPack {
         uint16_t value;
         struct {
             uint16_t        // total size
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -1,16 +1,13 @@
 #ifndef _EVT_GPIO_
 #define _EVT_GPIO_
 
-#include "pin.hpp"
-
-#include <stdint.h>
+#include <cstdint>
+#include <core/io/pin.hpp>
 
 namespace core::io {
 // Forward declarations:
 // The different pins are hardware specific. Forward declaration to allow
 // at compilation time the decision of which pins should be used.
-enum class Pin : uint8_t;
-enum class Port : uint8_t;
 
 /**
  * Interface for interacting with GPIO pins on a device. GPIO pins can have
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -1,6 +1,8 @@
 #ifndef _EVT_GPIO_
 #define _EVT_GPIO_
 
+#include "pin.hpp"
+
 #include <stdint.h>
 
 namespace core::io {
@@ -48,6 +50,7 @@ public:
      * then
      * a.pin_12 && a.pin_8 && a.pin_4 && a.pin_0 == 1
      */
+    #pragma pack(push, 1)
     union PinPack {
         uint16_t value;
         struct {
@@ -123,6 +126,10 @@ public:
      */
     GPIO(Pin pin, Direction direction, Pull pull = Pull::PULL_DOWN);
 
+    static uint16_t setPackBit(Pin pin) {
+        return 1 << pinNumberFromPin(pin);
+    }
+
     /**
      * Fill a given PinPack based on a given array of pins
      *
@@ -133,7 +140,7 @@ public:
     static void fillPinPack(PinPack& pp, Pin* pins, uint32_t num_pins) {
         pp.value = 0;
         for (uint32_t i = 0; i < num_pins; i++) {
-            pp.value |= 1 << (static_cast<uint16_t>(pins[i]) & 0x0F);
+            pp.value |= setPackBit(pins[i]);
         }
     }
 
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -9,7 +9,7 @@ namespace core::io {
 // Forward declarations:
 // The different pins are hardware specific. Forward declaration to allow
 // at compilation time the decision of which pins should be used.
-enum class Pin  : uint8_t;
+enum class Pin : uint8_t;
 enum class Port : uint8_t;
 
 /**
@@ -26,31 +26,31 @@ enum class Port : uint8_t;
  */
 class GPIO {
 public:
-    /**
-     * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
-     * the value is meant to quickly set or read a value with ease.
-     * the struct is meant for individually flipping bits in the field with a friendly format
-     * Meant for mass GPIO instantiation
-     *
-     * Usage Example 1:
-     * PinPack a = {.value = 0x0010};
-     * PinPack b = {.pin_4 = 1};
-     * then
-     * a == b
-     *
-     * Usage Example 2:
-     * PinPack a = {.pin_4 = 1, .pin_15 = 1};
-     * then reading from a.value gives
-     * a.value == 0x8010;
-     *
-     * Usage Example 3:
-     * uint16_t any_number = 0x1111;
-     * PinPack a;
-     * a.value = any_number;
-     * then
-     * a.pin_12 && a.pin_8 && a.pin_4 && a.pin_0 == 1
-     */
-    #pragma pack(push, 1)
+/**
+ * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
+ * the value is meant to quickly set or read a value with ease.
+ * the struct is meant for individually flipping bits in the field with a friendly format
+ * Meant for mass GPIO instantiation
+ *
+ * Usage Example 1:
+ * PinPack a = {.value = 0x0010};
+ * PinPack b = {.pin_4 = 1};
+ * then
+ * a == b
+ *
+ * Usage Example 2:
+ * PinPack a = {.pin_4 = 1, .pin_15 = 1};
+ * then reading from a.value gives
+ * a.value == 0x8010;
+ *
+ * Usage Example 3:
+ * uint16_t any_number = 0x1111;
+ * PinPack a;
+ * a.value = any_number;
+ * then
+ * a.pin_12 && a.pin_8 && a.pin_4 && a.pin_0 == 1
+ */
+#pragma pack(push, 1)
     union PinPack {
         uint16_t value;
         struct {
--- a/include/core/io/GPIO.hpp
+++ b/include/core/io/GPIO.hpp
@@ -26,38 +26,51 @@ enum class Port : uint8_t;
  */
 class GPIO {
 public:
-/**
- * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
- * the value is meant to quickly set or read a value with ease.
- * the struct is meant for individually flipping bits in the field with a friendly format
- * Meant for mass GPIO instantiation
- *
- * Usage Example 1:
- * PinPack a = {.value = 0x0010};
- * PinPack b = {.pin_4 = 1};
- * then
- * a == b
- *
- * Usage Example 2:
- * PinPack a = {.pin_4 = 1, .pin_15 = 1};
- * then reading from a.value gives
- * a.value == 0x8010;
- *
- * Usage Example 3:
- * uint16_t any_number = 0x1111;
- * PinPack a;
- * a.value = any_number;
- * then
- * a.pin_12 && a.pin_8 && a.pin_4 && a.pin_0 == 1
- */
-#pragma pack(push, 1)
+    #pragma pack(push, 1)
+    /**
+     * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
+     * the value is meant to quickly set or read a value with ease.
+     * the struct is meant for individually flipping bits in the field with a friendly format
+     * Meant for mass GPIO instantiation
+     *
+     * Usage Example 1:
+     * PinPack a = {.value = 0x0010};
+     * PinPack b = {.pin_4 = 1};
+     * then
+     * a == b
+     *
+     * Usage Example 2:
+     * PinPack a = {.pin_4 = 1, .pin_15 = 1};
+     * then reading from a.value gives
+     * a.value == 0x8010;
+     *
+     * Usage Example 3:
+     * uint16_t any_number = 0x1111;
+     * PinPack a;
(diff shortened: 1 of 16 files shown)


DannyCato:
Quick little rebase to not have previous commits of SDRAM stuff included where not need be

DannyCato:
This is where Initialization by port is added. I am now going to deprecate the older init state function and see to it from there

include/core/io/platform/f4xx/GPIOf4xx.hpp
```diff
@@ -47,7 +48,44 @@ class GPIOf4xx : public GPIO {
...
     static void gpioStateInit(GPIO_InitTypeDef* targetGpio, Pin* pins, uint8_t numOfPins, uint32_t mode, uint32_t pull,
-                              uint32_t speed, uint8_t alternate = 0x0DU);
+                              uint32_t speed, uint8_t alternate = 0x0E);
+
+    /**
+     * Initialize an arbitrary length array of GPIO pins
+     * @param pins Pins to set
+     * @param numOfPins number of pins in total
+     * @param mode gpio configuration mode
+     * @param pull pull-up or pull-down activation
```
aclowmclaughlin: what possible values for pull are there? Should it be a uint32, or would an enum be better?

include/core/io/platform/f4xx/GPIOf4xx.hpp
```diff
@@ -47,7 +48,44 @@ class GPIOf4xx : public GPIO {
...
+     * @param speed maximum gpio output frequency
+     * Possible values for Mode, Pull, and Speed can be found in "stm32f4xx_hal_gpio.h"
```
aclowmclaughlin, line 61: If there is a limited set of possible values, I'd prefer them to exist in this file. Preferable, in the function description itself, but if that's too unweildy, at least in a comment.

include/core/io/platform/f4xx/GPIOf4xx.hpp
```diff
@@ -47,7 +48,44 @@ class GPIOf4xx : public GPIO {
...
+     * Mass Initialize GPIO pins with the same value with a grouping by port
+     * @param pack_pins Byte Array of pins
+     * @param port Port of the pins to initialize
+     * @param mode gpio configuration mode
+     * @param pull pull-up or pull-down activation
+     * @param speed maximum gpio output frequency
+     * Possible values for Mode, Pull, and Speed can be found in "stm32f4xx_hal_gpio.h"
+     * @param alternate gpio alternate function selection
+     */
+    static void gpioPortInit(PinPack& pack_pins, Port port, uint32_t mode, uint32_t pull, uint32_t speed,
```
aclowmclaughlin: This doesn't need to be a reference since the Pinpack is a uint16 union, we could just pass a copy

include/core/io/GPIO.hpp
```diff
@@ -23,6 +26,54 @@ enum class Pin;
  */
 class GPIO {
 public:
+#pragma pack(push, 1)
```
aclowmclaughlin: What does the pragma do?

include/core/io/GPIO.hpp
```diff
@@ -23,6 +26,54 @@ enum class Pin;
...
+    /**
+     * A union of a 16-bit number and a  breakdown of it in a struct of individual bits.
+     * the value is meant to quickly set or read a value with ease.
+     * the struct is meant for individually flipping bits in the field with a friendly format
+     * Meant for mass GPIO instantiation
+     *
```
aclowmclaughlin, line 27: Very detailed description! Love it :3

include/core/io/pin.hpp
```diff
@@ -34,9 +37,9 @@ namespace core::io {
  * the pin is on and the number of the pin. We referenced MBed's documentation for generating
  * these values.
  */
-enum class Pin {
-    INVALID = -1, // THIS INTENTIONALLY DOES NOT POINT TO A PIN. Used as a default value, so the default value is
-                  // no longer PA_O (a real pin)
+enum class Pin : uint8_t {
+    INVALID = 0xFF, // THIS INTENTIONALLY DOES NOT POINT TO A PIN. Used as a default value, so the default value is
```
aclowmclaughlin, line 42: fyi -1 and FF will be the same thing but either works. This is a little more explicit what the actual value is.

include/core/io/pin.hpp
```diff
@@ -287,5 +290,46 @@ enum class Pin {
...
+};
+
+/**
+ * Pull the port from a Pin
+ *
+ * @param pin Pin
+ * @return Port of the Pin
+ */
+constexpr Port portFromPin(Pin pin) {
+    return static_cast<Port>((static_cast<uint32_t>(pin) & 0xF0) >> 4);
```
aclowmclaughlin, line 322: Can't you just left shift 4? (without masking?)

include/core/io/UART.hpp
```diff
@@ -13,7 +13,7 @@ namespace core::io {
 // Forward declarations:
 // The different pins are hardware specific. Forward declarations to allow
 // at compilation time the decision of which pins should be used.
-enum class Pin;
+enum class Pin : uint8_t;
```
aclowmclaughlin: Do we really need forward declarations? Shouldn't just including the pin.hpp class accomplish the same thing?
I know you didn't start this but I was wondering if you had any insight as to why this might be necessary?

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -178,13 +175,144 @@ void GPIOf4xx::registerIRQ(TriggerEdge edge, void (*irqHandler)(GPIO* pin, void*
...
+        break;
+#endif
+    default:
+        break; // Bad Port input
+    }
+}
+
+// do easy swaps without needing stack or calling additional function
+#define SWAP(A, B) \
+    do {           \
```
aclowmclaughlin: why is this in a do-while loop?
Also, xor might simplify this a little, and reduce the risk of overflow.
See https://en.wikipedia.org/wiki/XOR_swap_algorithm

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -1,3 +1,4 @@
+#include <algorithm>
```
aclowmclaughlin: What does this include do?

aclowmclaughlin: We might want to be careful if it is a C++ std library include- they tend to be ginormous

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -60,42 +59,36 @@ namespace core::io {
...
         break;
-    case 0x3:
+    case Port::D:
         this->port = GPIOD;
         break;
-    case 0x5:
+    case Port::F:
         this->port = GPIOF;
         break;
 #ifdef STM32F469xx // and not STM32F446
```
aclowmclaughlin: These ought to be switched over to using the HAS_PORT ifdefs, right?

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -178,13 +175,144 @@ void GPIOf4xx::registerIRQ(TriggerEdge edge, void (*irqHandler)(GPIO* pin, void*
...
+        break;
+    case Port::D:
+        __HAL_RCC_GPIOD_CLK_ENABLE();
+        HAL_GPIO_Init(GPIOD, &targetGpio);
+        break;
+    case Port::F:
+        __HAL_RCC_GPIOF_CLK_ENABLE();
+        HAL_GPIO_Init(GPIOF, &targetGpio);
+        break;
+#ifdef STM32F469xx // STM32F469xx only
```
aclowmclaughlin: Same thing, shouldn't we use the HAS_PORT ifdefs?

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -178,13 +175,144 @@ void GPIOf4xx::registerIRQ(TriggerEdge edge, void (*irqHandler)(GPIO* pin, void*
...
+    if (numOfPins == 0) {
+        return;
+    }
+    if (numOfPins == 1) {
+        gpioSingleInit(pins[0], mode, pull, speed, alternate);
+        return;
+    }
+
+    // HeapSort on Pin* pins from https://en.wikipedia.org/wiki/Heapsort#Standard_implementation
+    auto* sorted  = reinterpret_cast<uint8_t*>(pins);
```
aclowmclaughlin, line 252: I'm wondering if this heapsort can be
1. moved to it's own function
and 
2. be a constexpr so it can possibly be evaluated at compile time?

Also, I don't recall if you do, but it really should be explicitly indicated that the Pin array provided is sorted when the method is run, because it will remain sorted after the method is finished and this could lead to unexpected behaviour

aclowmclaughlin: Not sure if making it a constexpr would work, because I think any constexpr called by a non-constexpr function just is evaluated at runtime, but I'm not entirely sure. Could be cool to look into.

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -178,13 +175,144 @@ void GPIOf4xx::registerIRQ(TriggerEdge edge, void (*irqHandler)(GPIO* pin, void*
...
+    targetGpio.Pull      = pull;
+    targetGpio.Speed     = speed;
+    targetGpio.Alternate = alternate;
+
+    initHALGPIO(targetGpio, port);
+}
+
 void GPIOf4xx::gpioStateInit(GPIO_InitTypeDef* targetGpio, Pin* pins, uint8_t numOfPins, uint32_t mode, uint32_t pull,
                              uint32_t speed, uint8_t alternate) {
+#pragma deprecated("GPIOf4xx::gpioStateInit is deprecated, but available for backwards compatibility")
```
aclowmclaughlin: Suggest what function to use instead.

aclowmclaughlin (changes requested):
Amazing first draft.

I love so much of this, and its a very needed rework.

include/core/io/pin.hpp
DannyCato: it would only build if it was switched from 0xFF since technically -1 is outside the range of a uint8

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: This was an accidental include when writing "sorted" and has been removed

include/core/io/pin.hpp
DannyCato: Yes, but this is more explict about what's actually happening. I think the extra clock cycle cost is worth it so that people will know exactly what region of the original pin number we are interested in

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: I was thinking about moving it to its own function, but I wanted to avoid having out of place methods within the GPIOF4xx.cpp file, and thought making an entirely new file for a relatively small iterative sort was unnecessary.

The constexpr idea is interesting and I will do more research on that

include/core/io/pin.hpp
aclowmclaughlin: tbh it might not even take an extra instruction, I'm not sure if the compiler is smart enough to optimize it out or not.

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: It's not really necessary in this case, but it is good practice when you want to separate the macro from the outer scope. I will remove it since it does not help here.

The XOR swap looks interesting and I will implement it

include/core/io/UART.hpp
DannyCato: No, I switched them to simply including pin.hpp. Seems to me like that makes a lot more sense

include/core/io/GPIO.hpp
DannyCato: This ensures that no padding is placed in/around the struct for the CPU to have an easier time accessing the elements. I am not certain that it is needed, but I saw a handful of examples with them

include/core/io/platform/f4xx/GPIOf4xx.hpp
DannyCato: That makes sense, fixed

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: Yes, it has been changed to this

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: Change made

include/core/io/platform/f4xx/GPIOf4xx.hpp
DannyCato: There is a defined set of values, but it is rather extensive and contains more bit-packed information.

I think the comment directing people to look at stm32f4xx_hal_gpio.h works in this case. I think directing people elsewhere to look at the set of values is acceptable

include/core/io/platform/f4xx/GPIOf4xx.hpp
DannyCato: Honestly, we should enum it more the I think about it

include/core/io/platform/f4xx/GPIOf4xx.hpp
DannyCato: I'm turning it into an enum

DannyCato: I've tried to play with this for a long time, and it does not feel right. Directing people to the header file will remain the way I want to do it

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: Suggested!

DiegoLHendrix (approved):
LGTM

include/core/io/GPIO.hpp
aclowmclaughlin: I think there's a gcc attribute that does the same? I'm not sure which is better to use.

src/core/io/platform/f4xx/GPIOf4xx.cpp
```diff
@@ -6,6 +6,7 @@
 #include <HALf4/stm32f4xx_hal_gpio.h>
 #include <HALf4/stm32f4xx_hal_rcc.h>
 #include <core/platform/f4xx/stm32f4xx.hpp>
+#include <sys/types.h>
```
aclowmclaughlin: What are we using this for?

aclowmclaughlin (changes requested):
Looking good, I like the progress. Still some stuff to clean up.

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: Idk, removed

include/core/io/GPIO.hpp
DannyCato: switched to it since it can more directly be related to the union

src/core/io/platform/f4xx/GPIOf4xx.cpp
DannyCato: From the research I've done, it is possible, but will be template hell. Basically need to make a constexpr variable that takes in the input of a constexpr function with a parameter of a template struct that mimics how std::array works to get it working at compile time.

I want to get working on hudl so that will be a problem for the next guy

aclowmclaughlin (approved):
Looks good. Please test a couple of samples to make sure any current projects that use GPIO work, then feel free to merge.

ActuallyTaylor (approved):
Looks great!

PR #143, opened 2026-04-19
For what is lost in a listing? It seems to me as though the truth that comes from any listing whatsoever is the fact that it has to come from somewhere and naturally will always be in the last place that you look. However, the only thing left to do is search and hopefully find. For you do not know where something might be found until you take the time to go looking for it. The joke is that listing files are really long (like the title of this pr).

DannyCato:
The only thing that this does is change a small part of the EVT-Core CMake file for compiling to specify that after it is done it should pipe the output to a file called build/{Project_Name}.lst

### DIFF
--- a/cmake/evt-core_build.cmake
+++ b/cmake/evt-core_build.cmake
@@ -1,11 +1,12 @@
 #[[
-Creates an executable with the provided name. Generates bin, hex, elf, and map
-files for the given name. This will make a dedicated project.
+Creates an executable with the provided name. Generates bin, hex, elf, lst and map
+files for the given name. This will compile a dedicated project from the passed in source files.
 ]]#
 macro(make_exe proj_name sources)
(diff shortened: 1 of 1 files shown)


ActuallyTaylor:
I love the incredibly long PR title with no description

ActuallyTaylor (approved):
Looks good! Out of curiosity, why do we need this?

aclowmclaughlin (approved):
Looks good. Add a description and use a different name for the merge though.

DannyCato:
> Looks good! Out of curiosity, why do we need this?

It's helpful for seeing what the compiler is actually doing with our code. It will take the C/C++ line by line and show what the compiler makes as output assembly. It's just another tool we can use to see what is happening

DannyCato:
Also, description has been added

PR #146, opened 2026-09-15
CAN we be clear?

ActuallyTaylor:
A super small PR that adds a comment explaining why the NODE-ID is not added into the TPDO’s settings object.


include/core/io/CANOpenMacros.hpp
```diff
@@ -236,42 +236,44 @@
  * This macro creates an TPDO settings object. This macro itself is abstract,
  * allowing it to be used with any TPDO number supported by CANOpen.
  *
+ * Note: The TPDO Settings object's TPDO_ID is incremented by the NODE-ID within the CANopen driver.
+ *
  * @param TPDO_NUMBER (integer) the TPDO number this settings object is for.
  * @param TRANSMISSION_TYPE (hex) the type of transmission to make. You should use TRANSMIT_PDO_TRIGGER_TIMER.
```
tmb5932, line 242: What are your thought about renaming (or adding a new define with same value for backwards compatibility) the `TRANSMIT_PDO_TRIGGER_TIMER` because I think calling it a timer based trigger is a bit confusing when it is also used if you want to do software-driven-tpdo-sending (with AlertTPDO()).

I was thinking that calling it `TRANSMIT_PDO_TRIGGER_EVENT` or something so it can cover timers and software defined triggers?

Again if u have a reason / argument against it, lmk. It’s something that I didn’t think it was a big enough deal when I originally added alertTPDO, but now that we’re editing the exact file, I thought I might as well mention it.

tmb5932 (approved):
Looks good, but read my comment n let me know what u think

include/core/io/CANOpenMacros.hpp
ActuallyTaylor: This is new knowledge to me, in our existing CAN documentation timer trigger is the only thing mentioned. I can't find any documentation on `AlertTPDO`, do you happen to have a link to it?

I think changing it to `TRANSMIT_PDO_TRIGGER_EVENT` is a good option to make it clearer.

tmb5932: Sorry, my comment was unclear, and any missing EVT documentation is my fault.

I added support to EVT-core a while ago to do a “Trigger PDO” API call through an EVT-Core function called `alertTPDO()`, as I needed it for VCU's health status canopen messages.

I'm on phone so I can't link figure out how to url or a line but my code I added is at `include/core/io/CANopen.hpp` on line 106

The actual api function is `COTPdoTrigPdo()`. You can read about it here https://canopen-stack.org/v4.2/api/tpdo/

DannyCato (approved):
LGTM

include/core/io/CANOpenMacros.hpp
DannyCato: I think it is always be better to name things more clearly. If we could get an overlapping name so that we do not need to be concerned about future confusion, then we would be pretty perfect.

ActuallyTaylor: @tmb5932 That sounds good, but I think we should take another approach based on what @DannyCato said. 

I want to propose a new macro that uses `0xFF` which is accepted as an asynchronous transmit trigger according to the [EVT Canopen](https://docs.google.com/document/d/1T2wHcIqxxq6p3vnUJfuLJ6D8pbDnZn5vJuaJZwCx5HE/edit?tab=t.0) google doc we have. This allows us to keep a logical separation for quickly identifying how a dictionary transmits while providing a better interface for the alert events.

Let me know what you all think!

```cpp
#define TRANSMIT_PDO_ALERT_EVENT 0xFF
```

tmb5932: Yeah that looks good. Forgot that both mean the same but have diff implications and restrictions (that I agree we can ignore because it's only used in house and not sold).

I don't think we need to worry about CiA regulations for this kind of stuff honestly. (With FF technically having regulations by CiA committee on what counts as a valid event to trigger a transmit on) 

Who's gon check us lol.

ActuallyTaylor: Wait they do? Cause that is news to me 😭. Maybe we just add a comment to note this and if we ever do need CiA regulations we can change it back to both be `0xFE`. Do you have the link for the regulations?

DannyCato: No one is going to stop us from testing the bike because of this (knock on wood)

tmb5932: I added the CiA application layer documentation that defines it to the shared drive (you need an account or view it on their website).

The 0xFE vs 0xFF is mentioned on `page 138` (and maybe other places I just quick scanned through)

It can be found here: 

https://drive.google.com/file/d/1hzFuT4gDF0-7ZhtvNa2hiTHNDBtrgcvc/view?usp=drivesdk

ActuallyTaylor: Sounds good, it looks like OxFF falls into application specific which should clear us to use it even under the spec.

tmb5932: Are you saying that as in you looked at the CiA profile specs and found one that works for us or are you thinking application profile means the app chooses it?

Here, `application profile` means a published CANopen system-profile specification, like cia 401, 402, etc

Now, I haven't actually looked at any of them individually, but I do know that they define what constitutes an event that is allowed (or not allowed) to trigger a TPDO;
They also define when received RPDO data takes effect; 
And stuff like that.

ActuallyTaylor: Ohh that makes more sense. Yeah that was confusion on my end. I thought `application profile` meant it was up to us as the application to make that choice.

tmb5932: Yeah, it's super confusing idk why they wouldn't name it something else. 

But honestly, I still agree that we should add it and fuck cia. It's meant for products and things unrelated people interface with, which is not this

